PR #7, opened 2026-03-29
fix(install): avoid duplicate PATH entry when already present

voronkovich:
This change adds an early check against the live `PATH` variable. If `$INSTALL_DIR` is already present, the function reports success and returns immediately, avoiding unnecessary rc file modifications and duplicate entries.

In most cases `.local/bin` path is already configured by user, so we don't need to change user's rc file at all.


geodro:
Thanks!

PR #212, opened 2026-04-17
feat(tui): add lerd tui terminal dashboard

geodro:
## Summary

Adds `lerd tui`, a btop-style terminal dashboard with near parity to the existing web UI and system tray. Built on the same bubbletea / lipgloss stack used for `lerd man` and the same `siteinfo` / `podman.Cache` / eventbus plumbing that drives `lerd-ui`, so both surfaces see identical live state.

- **Layout**: sites + services stacked left, full-height site detail right, toggleable log pane below with a right-edge scrollbar, header with DNS / nginx / FPM status and a cached `update: vX.Y.Z` banner.
- **Site detail**: primary domain header, internal name, disk path, full domains list (add with `a`, rename with `e`, remove with `x`), services used with live state, workers with toggles, git worktrees, HTTPS toggle, LAN share toggle with full `http://<lan-ip>:<port>` URL, inline PHP and Node version pickers backed by `lerd isolate` / `lerd isolate:node`.
- **Services pane** includes site-owned workers (`queue-<site>`, `schedule-<site>`, `horizon-<site>`, `reverb-<site>`, custom framework workers) routed to the right per-site commands.
- **Filter + sort** on both lists (`/` filter, `o` cycle sort). `v` hides services, `t` opens a shell in the focused container, `l` tails logs, `[` / `]` cycle log sources (FPM + custom container, worker journals via `journalctl --user` on Linux / `podman logs` on macOS, framework app log files tailed with `tail -F`).
- **In-pane overlays** for global Settings (`S`) and a Keybindings reference (`?`), both toggleable back with `esc`.
- **Cross-platform**: platform-split for worker log tailing so the TUI works on macOS (where workers are containers) and Linux (where they are systemd user units).
- Live updates via the in-process eventbus plus a 2s refresh tick so changes made from another terminal surface within a couple of seconds.
- Unit tests for filter / sort variants, `detailRows`, log targets, worker-row synthesis, `nextFocus` edges, domain input flow, picker lifecycle, and row alignment invariants.

Docs: new `/features/tui.md` page with a full keybindings reference, updated homepage feature tile, README feature bullet and comparison table row, changelog entry under Unreleased. Post-install banner now also points at `lerd tui` alongside the dashboard URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ If you're a PHP developer on Linux and want frictionless local development — a
 - 🐘 **Per-project PHP version** (8.1–8.5), switch with one click
 - 📦 **Node.js isolation** per project (Node 22, 24)
 - 🖥️ **Built-in Web UI** - 3-pane dashboard to manage sites, services, and logs from a browser
+- 🖥️ **Terminal dashboard** (`lerd tui`) - btop-style TUI with live status, site detail pane, inline domain and version editing, shell drop-in, log tailing, and filter/sort — the same operations surface as the web UI, for tmux and SSH workflows
 - 🗄️ **One-click services**: MySQL, PostgreSQL, Redis, Meilisearch, RustFS, Mailpit, Stripe Mock, Reverb and more
 - 📋 **Live logs** for PHP-FPM, Queue, Schedule, Reverb, per site
 - 🔒 **Rootless & daemonless** - Podman-native, no Docker required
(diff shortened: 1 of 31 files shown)


geodro:
continuation of closed #211 (previous base branch was deleted, which auto-closed it). all commits preserved.

geodro:
closing in favour of reopened #211 which preserves the original history.

PR #213, opened 2026-04-17
fix(worktree): copy vendor and node_modules instead of symlinking

### LINKED ISSUE #208: Weird autoload behavior in worktrees
Hey, I've seen some weird things with my repo when using worktrees that I wanted to talk to you about.

Currently, when the watcher detects a new worktree, it creates symlinks to the vendor and node_modules directories in the worktree. This makes sense on the surface because it saves a bunch of disk space, but it also seems to be problematic in my case. I'm not sure whether this is specific to me or more global.

In my vendor/ directory, I have an `autoload.php` from composer, which ends up initialising the `ClassLoader` like this:
```php
self::$loader = $loader = new \Composer\Autoload\ClassLoader(\dirname(__DIR__));
```

The problem is that `__DIR__` refers to the main git repo directory, not the worktree, so classes end up being loaded from the main repo too. You can see a concrete example on this image from PhpStorm:

Image

As a sanity check I tried an `echo __DIR__;` in the `autoload.php`, and it's confirmed below in my terminal output:
```
┌─[ropi][framework][±][t3code/mcp-entries-list U:4 ?:3 ✗][~/.../brokerchooser.com/t3code-d3fe6444]
└─▪ pwd
/home/ropi/.t3/worktrees/brokerchooser.com/t3code-d3fe6444
┌─[ropi][framework][±][t3code/mcp-entries-list U:4 ?:3 ✗][~/.../brokerchooser.com/t3code-d3fe6444]
└─▪ cd vendor/
/home/ropi/.t3/worktrees/brokerchooser.com/t3code-d3fe6444/vendor
┌─[ropi][framework][±][master U:1 ?:2 ✗][~/.../t3code-d3fe6444/vendor]
└─▪ cd ..
┌─[ropi][framework][±][t3code/mcp-entries-list U:4 ?:3 ✗][~/.../brokerchooser.com/t3code-d3fe6444]
└─▪ lerd php artisan route:clear
Xdebug: [Step Debug] Could not connect to debugging client. Tried: host.containers.internal:9003 (through xdebug.client_host/xdebug.client_port).
/home/ropi/Projects/brokerchooser.com/vendor
   INFO  Route cache cleared successfully.

┌─[ropi][framework][±][t3code/mcp-entries-list U:4 ?:3 ✗][~/.../brokerchooser.com/t3code-d3fe6444]
└─▪
```

I did check, just to make sure, that `cd`ing into the vendor directory keeps `pwd` in the worktree, but it looks like PHP disregards this. What do you think would be the best solution here? Is it something specific to me? (I'm not that long into the PHP ecosystem, I've been more of a Java guy most of my career)

geodro:
## Summary
PHP resolves `__DIR__` through symlinks, so `vendor/autoload.php` inside a git worktree reports the main repo's real path. Composer's `ClassLoader` then initialises with the main repo as its base and every PSR-4 class resolves against main's source tree instead of the worktree's. Any worktree with diverged `app/`, `src/`, etc. silently runs stale code. Reported in #208.

Replace the symlinks with real copies seeded from the main repo, using `cp -a --reflink=auto` on Linux and `cp -Rc` on macOS (APFS clone_file) for near-instant, zero-disk copies on btrfs, xfs-reflink, and APFS. Falls back to a plain Go walk on ext4 and anywhere `cp` is unavailable.

After copying, the JS install is routed to the matching package manager based on the lockfile (pnpm-lock.yaml → pnpm, yarn.lock → yarn, bun.lock / bun.lockb → bun, package-lock.json / npm-shrinkwrap.json → npm ci, otherwise npm install). composer install goes through the lerd shim; the JS manager goes through the shim when it's npm and through PATH otherwise. Missing binaries aggregate into a warning rather than failing the worktree so the reflink copy is still usable.

Legacy symlinks from earlier lerd versions are detected on the next worktree detection event and replaced with real copies automatically.

Fixes #208.

### DIFF
--- a/docs/features/git-worktrees.md
+++ b/docs/features/git-worktrees.md
@@ -36,11 +36,17 @@ When a worktree vhost is first created, Lerd sets up three things in the checkou
 
 | Resource | Behaviour |
 |---|---|
-| `vendor/` | Symlinked from the main repo (shares Composer packages) |
-| `node_modules/` | Symlinked from the main repo (shares npm packages) |
+| `vendor/` | Copied from the main repo using reflinks where the filesystem supports them (btrfs, xfs-reflink, APFS), then reconciled against the worktree's own `composer.lock` via `composer install` |
+| `node_modules/` | Copied from the main repo (reflink where supported), then reconciled against the worktree's own `package-lock.json` via `npm ci` |
 | `.env` | Copied from the main repo with `APP_URL` rewritten to `http://<branch>.<site>.test` |
 
-If any of these already exist in the worktree they are left untouched.
+If `vendor/` or `node_modules/` already exist as real directories they are left untouched. Legacy symlinks left by earlier lerd versions are replaced with real copies.
+
+On reflink-capable filesystems the initial copy is near-instant and consumes no extra disk until a file is modified. On ext4 it falls back to a plain copy, which typically takes a few seconds per vendor tree. The subsequent `composer install` / `npm ci` is a quick verification pass when the worktree branch shares the same lockfile as main, and installs only the differences when the branch has changed dependencies.
+
+::: info Why not symlink?
+Earlier versions of lerd symlinked `vendor/` to save disk. PHP resolves `__DIR__` through symlinks to the real filesystem path, so Composer's `ClassLoader` would initialise against the main repo directory and silently load stale classes from there. Real copies avoid the problem while still being cheap on modern filesystems.
+:::
 
 ---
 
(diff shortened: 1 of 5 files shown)


geodro:
continuation of closed #209 (previous base branch was deleted, which auto-closed it). all commits preserved including the pnpm, yarn, bun detection fix added after @ropi-bc's feedback.

geodro:
closing in favour of reopened #209 which preserves the original history.

PR #238, opened 2026-04-22
fix(bootstrap): make project_new + env_setup reach a working site without agent hand-holding

geodro:
Observed in a real Claude Code session: the agent scaffolded a Laravel project via `project_new`, called `site_link` + `env_setup`, and the site returned **500** on every request. It took three extra back-and-forth turns (`composer install` → `key:generate` → `touch database/database.sqlite` → `artisan migrate`) to get to "Laravel welcome page loads". Every step that failed is something lerd already knew about; the failures were all because the tools stopped short of the obvious next action.

## Root causes

**1. `project_new` left `vendor/` empty.** Framework `create` commands use `--no-install --no-plugins --no-scripts` (deliberate — keeps scaffold deterministic), so `project_new` returned "success" with no dependencies installed. Agents have no reason to know this.

**2. `env_setup` skipped the sqlite-file-creation block.** `env_setup` runs `lerd env`. For Laravel's default `DB_CONNECTION=sqlite`, the DB-swap prompt in `env.go` is gated behind `isInteractive()`. Via MCP there is no TTY, the prompt is skipped, and the sqlite-file-creation block below (itself gated behind `lerdYAMLServices["sqlite"]`) never fires. Result: no `database/database.sqlite`, Laravel 500s immediately.

**3. SKILL.md workflows didn't cover end-to-end bootstrap, cloned project setup, or 500-debugging,** so agents had to improvise.

## Fixes

- `internal/cli/env.go`: non-interactive callers (MCP, scripts) default the sqlite choice to "keep sqlite" instead of skipping entirely. The choice is persisted to `.lerd.yaml` and the existing sqlite-file-creation block runs, so `database/database.sqlite` is touched and the site can respond. Agents can still call `db_set` / `lerd db set` to switch to mysql/postgres afterwards — nothing locks in sqlite beyond the default.
- `internal/mcp/server.go`: `execProjectNew` now runs `composer install --no-interaction` in the appropriate FPM container when `composer.json` is present but `vendor/` is missing. New `runComposerInstallIfNeeded` helper is unit-tested against both no-composer.json and vendor-already-present cases so happy paths don't regress into re-pulling on re-runs.
- `internal/cli/mcp.go` (SKILL.md): replaced the two Laravel-specific workflows with three framework-agnostic end-to-end flows:
  - **Bootstrap a new project from scratch** — uses `project_new`, shows the `artisan` vs `console` fork for the migrate step, notes the new composer-install-after-scaffold and DB-auto-create behaviour.
  - **Set up a cloned project** — `composer install` **before** `env_setup` so APP_KEY generation has a valid `vendor/`, then migrate.
  - **Debugging a 500** — ordered checklist starting at `logs()`, then `env_check` / `which` / composer install / APP_KEY / migrate / service start / `doctor` as last resort.

Bumped `claudeSkillContent` ceiling 41000 → 42000 to fit the new workflows; any further growth still requires an explicit ceiling bump so accretion stays visible in diffs.

## Net effect

Agents running `project_new` → `site_link` → `env_setup` now get a working site in the same four tool calls that previously failed. The clone workflow is symmetric. The debug flow turns "I'm stuck on a 500" into a deterministic sequence that any model (including weaker local LLMs) can follow.

### DIFF
--- a/internal/cli/env.go
+++ b/internal/cli/env.go
@@ -198,26 +198,31 @@ func runEnv(_ *cobra.Command, _ []string) error {
 	// hasn't yet picked a DB service for this project, offer to swap sqlite for
 	// a lerd-managed mysql/postgres. Skipped for frameworks with explicit env
 	// service rules (e.g. wordpress, symfony) — they don't use DB_CONNECTION.
-	// "Keep SQLite" is persisted by adding "sqlite" to the project's services
-	// list so we don't re-ask on every subsequent `lerd env` run.
-	if len(fw.Env.Services) == 0 && isInteractive() &&
+	// Non-interactive callers (MCP, scripts) fall through to sqlite by default
+	// so they don't hit a 500 from the missing .sqlite file; the user can
+	// still switch later with `lerd db set mysql` or the db_set MCP tool.
+	if len(fw.Env.Services) == 0 &&
 		!lerdYAMLServices["mysql"] && !lerdYAMLServices["postgres"] && !lerdYAMLServices["sqlite"] &&
 		strings.EqualFold(strings.TrimSpace(envMap["DB_CONNECTION"]), "sqlite") {
 
 		dbChoice := "sqlite"
-		dbForm := huh.NewForm(huh.NewGroup(
-			huh.NewSelect[string]().
-				Title("Database").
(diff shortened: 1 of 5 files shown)


geodro:
Consolidated into single branch work/fix/agent-site-bootstrap locally for testing. Will reopen a single PR when ready.

PR #263, opened 2026-04-26
(fix): Align `.env` `APP_URL` with worktree domain every time a worktree is scanned

ropi-bc:
My workflow involves my tool creating a worktree with a random ID, and then renaming it later. This means that I always need to rewrite the `APP_URL` in my `.env` files manually. This PR makes it so that every time a worktree is scanned, even if it isn't new, the watcher makes sure the `APP_URL` is up-to-date.

I'd like to file a PR that triggers the watcher even when a branch/worktree is renamed.

PS.: I'm sorry about the mismatch between @rolaca11 and @ropi-bc. Both of them are my accounts. One is work, one is my personal. I'm trying to use both on the same machine, but I sometimes mix them up.


geodro:
Thanks for both this and 264, the rename pain is real and I want to land a fix. Having read 264 alongside this one, I think we should reshape what goes where.

The cleanupWorktreeVhosts change in this PR doesn't pay for itself once 264 lands. 264's syncWorktree path with pruneStale is the right place to handle renames, and it does the surgical thing of removing only the now stale vhost rather than nuking every vhost on the site. Calling EnsureWorktreeDeps from cleanupWorktreeVhosts also has a hidden cost, EnsureWorktreeDeps runs composer install and the JS install via InstallDependencies, so deleting one worktree would kick off a full deps install for every surviving worktree on the site. Could you drop that change from this PR?

The rewriteAppURL on existing files is the right behaviour and it's what makes 264 actually solve the stated problem. Without this, 264's HEAD watcher fires syncWorktree, which calls EnsureWorktreeDeps, which sees the .env already exists and returns without touching APP_URL. So we do need this part. Two small fixes I'd like before merging though. The function writes os.WriteFile with mode 0644 unconditionally, which silently resets anyone who chmod 600'd their .env, the file holds APP_KEY and DB creds and 0600 is the right thing to do there. Could you stat first and preserve the existing mode? Also, since the function now runs on every sync and not just first creation, please short circuit when the new bytes match the existing bytes, otherwise mtime bumps on every call and any external watcher on .env will see a constant stream of changes.

Honestly, given that 263 and 264 only make sense together, I'd lean toward folding them into a single PR, the rewriteAppURL change here plus the HEAD watcher and syncWorktree refactor from 264. If you'd rather keep them separate that's fine, but in that case let's land 264 first and then this one with just the rewriteAppURL change and the perm and mtime fixes above.

rolaca11:
cool, I'll look at these comments soon. thank you for your time, I'll merge the 2 PRs together

PR #268, opened 2026-04-29
Gotenberg API for PDF generation and document conversion

codeduction:
Add preset for gotenberg service that is used for PDF generation, used in laravel-pdf for example.

### DIFF
--- a/internal/config/presets/gotenberg.yaml
+++ b/internal/config/presets/gotenberg.yaml
@@ -1,0 +1,13 @@
+name: gotenberg
+image: docker.io/gotenberg/gotenberg:8
+ports:
+  - 3000:3000
+environment:
+  API_ENABLE_BASIC_AUTH: "true"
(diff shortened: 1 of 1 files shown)


geodro:
thanks for the gotenberg preset, really nice addition for laravel pdf projects. merging now and pushing a small follow up since gotenberg v8 reads API_BASIC_AUTH_USERNAME and API_BASIC_AUTH_PASSWORD without the GOTENBERG_ prefix so the auth env vars probably are not getting picked up, plus our other presets like mailpit and mongo express skip auth entirely since the lerd network already isolates the container. while in there i will quote the port to match the rest of the presets, add update_strategy rolling so the dashboard surfaces v8 updates, and add env_detect on spatie/laravel-pdf so it auto suggests for laravel apps that pull that package in. credit stays on your commit, appreciate the contribution.

PR #303, opened 2026-05-07
fix: MariaDB client compatibility with MySQL 8.4 in PHP-FPM containers

Bottelet:
## Summary

Lerd's PHP-FPM containers ship `mysql-client` from Alpine, which is the **MariaDB client**. When connecting to lerd's MySQL 8.4 service, three incompatibilities cause failures in CLI tools (`mysql`, `mysqldump`) that Laravel and packages shell out to:

1. **SSL rejection**: MySQL 8.4 auto-generates self-signed TLS certificates. The MariaDB client verifies the certificate chain and rejects them with `ERROR 2026 (HY000): TLS/SSL error: self-signed certificate in certificate chain`. This breaks Laravel's schema dump loading (`php artisan migrate` with `database/schema/*.sql`) and third-party tools like `spatie/db-dumper`.

2. **Auth plugin mismatch**: MySQL 8.4 defaults to `caching_sha2_password` and disables `mysql_native_password`. The MariaDB client doesn't bundle the `caching_sha2_password` plugin. Loading the plugin via `mysql-native-password=ON` isn't enough — the entrypoint still creates users with the 8.4 default. Adding `authentication_policy='mysql_native_password,,'` makes the entrypoint use native_password for `CREATE USER`.

3. **Strict FK enforcement**: MySQL 8.4 added `restrict_fk_on_non_standard_key` (defaults to ON), which rejects foreign keys referencing non-unique columns. This breaks existing schema dumps that worked on MySQL 8.0.

> **Note:** PHP's PDO driver handles both SSL and `caching_sha2_password` natively, so Eloquent queries work fine. The failures are specifically in CLI tools that Laravel and packages shell out to.

## Changes

- **PHP-FPM Containerfile**: Add `/etc/my.cnf.d/lerd-no-ssl.cnf` with `[client] ssl=0` so the MariaDB CLI client skips SSL verification on the trusted lerd network. Also applied to the fast-path build for pre-built base images.
- **MySQL preset config**:
  - `loose-mysql-native-password=ON` — loads the plugin on startup
  - `authentication_policy='mysql_native_password,,'` — makes the entrypoint create users with native_password auth
  - `loose-restrict-fk-on-non-standard-key=OFF` — disables strict FK enforcement so existing schema dumps work on 8.4
- **FFmpeg**: Added `ffmpeg` to the PHP-FPM image — commonly needed by PHP media libraries (e.g. `spatie/media-library`) for video/audio processing.
- **Regression test**: Asserts the three MySQL 8.4 compat directives are present in the preset config.

## Upgrade note for existing MySQL 8.0 → 8.4 users

`authentication_policy` only affects newly created users. Users who upgrade an existing MySQL 8.0 install (stale-image migration auto-bumps to 8.4) will keep `root` on `caching_sha2_password`. They'll need either:
- `lerd service reinstall mysql --reset-data` (fresh database), or
- `podman exec lerd-mysql mysql -u root -plerd -e "ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY 'lerd';"`

## Test plan

- [x] `lerd service reinstall mysql` picks up all three directives in `/etc/mysql/conf.d/lerd.cnf`
- [x] `lerd php:rebuild <version>` produces an image with `/etc/my.cnf.d/lerd-no-ssl.cnf` and `ffprobe` available
- [x] `php artisan migrate:fresh --seed` works on a Laravel project with a `database/schema/mysql-schema.sql` dump
- [x] `mysqldump` from PHP-FPM container connects without SSL errors
- [x] Schema dumps with FK references to non-unique columns load without errors
- [x] `go test ./internal/config/ -run TestMySQLPresetContainsCompatDirectives` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


geodro:
thanks for putting this together, the SSL config and the FK option both look right and I verified them. the auth plugin part doesn't quite work yet on a fresh mysql 8.4 install though, I tested it with your exact preset_files.go config and root still comes up as caching_sha2_password, so the mariadb client still fails with "Plugin caching_sha2_password could not be loaded". loading the mysql_native_password plugin via mysql-native-password=ON makes it available but doesn't change the auth method that the entrypoint uses when creating users, so root keeps the 8.4 default.

the missing piece is authentication_policy='mysql_native_password,*,*' alongside the loose-mysql-native-password=ON line. once that's set, the entrypoint's CREATE USER picks up native_password and root comes up with the right plugin, I retested and the mariadb client connects fine. the loose- prefix isn't needed on authentication_policy since it's been around since 8.0.

while you're in there, could you also add a small test in internal/config/preset_files_test.go (or a new file) that asserts presetFiles["mysql"] contains the three directives we care about, mysql-native-password=ON, authentication_policy with native_password, and restrict-fk-on-non-standard-key=OFF, so we don't accidentally regress the my.cnf in a future edit. a one-shot string contains check is plenty, no need for anything fancy.

one more thing worth flagging in the description, the loose-restrict-fk-on-non-standard-key=OFF directive is in the diff but not mentioned in the PR body, would be good to add a sentence so it lands in the changelog.

also a note for users who upgrade an existing mysql:8.0 install, the stale-image migration auto-bumps them to 8.4, the new authentication_policy only affects newly created users so their existing root account keeps caching_sha2_password and the mariadb client still won't auth against it. they'd need either a reinstall with --reset-data or an ALTER USER 'root'@'%' IDENTIFIED WITH mysql_native_password BY 'lerd' run inside the container. happy to handle the release note for that part on our end.

Bottelet:
@geodro  thanks for the review! pushed the authentication_policy fix and added the test.

PR #308, opened 2026-05-07
feat: env_overrides in .lerd.yaml for worktree env templating

Bottelet:
## Summary

Adds `env_overrides` to `.lerd.yaml`, allowing projects to define templated or static env variables that are resolved when worktrees are created. This replaces the hardcoded `APP_URL=scheme://domain` rewrite with a flexible system that supports any env key.

## Problem

Multi-tenant apps (and other projects) often need env variables derived from the site domain — not just `APP_URL`. For example, a Laravel app with tenant subdomains needs:

```
APP_URL=https://app.feature-branch.myapp.test
CENTRAL_DOMAIN=feature-branch.myapp.test
```

Previously, lerd only rewrote `APP_URL` to `scheme://worktreeDomain`. Any other domain-dependent env vars kept the main repo's values, breaking tenant routing, session cookies, etc.

## Solution

A new `env_overrides` map in `.lerd.yaml`:

```yaml
env_overrides:
    APP_URL: "{{scheme}}://app.{{domain}}"
    CENTRAL_DOMAIN: "{{domain}}"
    CACHE_DRIVER: redis
```

Values can use template placeholders or be plain static strings. When a worktree is created, each value is resolved and written into its `.env`.

| Placeholder | Resolves to | Example |
|-------------|-------------|---------|
| `{{domain}}` | Worktree domain | `feature-branch.myapp.test` |
| `{{scheme}}` | `http` or `https` | `https` |
| `{{site}}` | Database-safe name (underscored) | `feature_branch_myapp_test` |

When `APP_URL` is present in overrides it takes precedence over the default rewrite. Without `env_overrides`, behavior is unchanged.


geodro (changes requested):
Nice addition, the env_overrides system is more flexible than the old APP_URL only rewrite, and the four new tests cover the placeholder resolution, static values, new key append, and fallback path. I pulled the branch and ran internal/git and internal/config locally, all green.

A couple of things before this lands. Main one is docs, since this adds a new public facing config key, docs/features/git-worktrees.md still describes the worktree env behavior as APP_URL only at lines 57, 70, and 88, and docs/features/project-setup.md doesn't mention env_overrides either. Could you add a short section with the placeholder table from the PR body and an example yaml block.

Two small ones, mind stripping the Generated with Claude Code footer from the PR body, and a rebase against main when you get a chance since the branch is behind.

Optional follow up, the {{site}} sanitization in applyEnvOverrides overlaps with similar logic in internal/cli/env.go, internal/cli/worktree_db.go, and internal/mcp/server.go. Could be worth extracting a shared config.SiteSlug helper at some point so the rules stay aligned, but no need to do it here.

Bottelet:
thanks for the review! added the env_overrides docs section to git-worktrees.md with the placeholder table and example yaml. footer stripped, branch is already on main.

good call on the SiteSlug extraction — the sanitization logic does repeat in a few places. happy to do that as a follow-up once this lands.

PR #309, opened 2026-05-07
feat: automatic wildcard cert SANs and nginx server_name for worktree subdomains

Bottelet:
## Summary

When a worktree is created on a secured site, lerd now automatically includes `*.branch.domain.test` in both the TLS certificate SANs and the nginx `server_name` directive. This means deep subdomains like `app.branch.domain.test` work out of the box without manual cert regeneration or nginx config edits.

## Problem

Multi-tenant apps often use subdomains of the site domain (e.g. `app.myapp.test`). When a worktree is created, the worktree domain becomes `branch.myapp.test`, and the tenant subdomain becomes `app.branch.myapp.test`. X.509 wildcards only go one level deep — `*.myapp.test` covers `branch.myapp.test` but not `app.branch.myapp.test`.

Previously, developers had to manually:
1. Regenerate the cert with the extra SAN
2. Update the nginx `server_name` to include the wildcard
3. Reload nginx

This had to be repeated for every new worktree.

## Solution

- **Nginx vhost**: Worktree vhosts now include `*.worktreeDomain` in `server_name` (both HTTP and SSL variants), so nginx accepts requests for any subdomain of the worktree
- **Certificate**: When a worktree is created on a secured site, the parent cert is reissued to include `worktreeDomain` + `*.worktreeDomain` as SANs alongside all existing domains
- **Exported `ReissueCertForWorktree`**: Available for the daemon sync flow to call after worktree creation

All changes are backwards-compatible — sites without worktrees are unaffected, and the cert reissuance only happens on secured sites.


geodro (changes requested):
Implementation looks good, the wildcard SAN does land in the cert via IssueCert's existing d plus *.d expansion, so deep subdomains like app.branch.myapp.test will validate. Build and the cert, nginx, and cmd test packages all pass locally on the branch.

Main blocker is docs, this PR actively changes the behavior described in docs/features/https.md and docs/features/git-worktrees.md. Both currently say lerd reuses the parent's wildcard cert with no extra issuance, but after this PR the cert is reissued on every worktree add to pick up *.branch.myapp.test, which is the whole point of the change. Could you update those sections to describe the reissue and the deep subdomain reason behind it.

Tests would be great here too, the new ReissueCertForWorktree and issueCertWithWorktrees helpers and the *.<domain> server_name change are uncovered. Even just an nginx test asserting the wildcard lands in the rendered vhost, plus a unit test on the domain list builder, would catch future regressions.

Two smaller follow ups, the cert is never reissued when a worktree is pruned so the SAN list grows over time, and there's a brief unsecured window between os.Remove and a failed IssueCert that the [WARN] log doesn't really surface. Both are low impact since this is mkcert, but worth either fixing here or filing as issues.

Same housekeeping as the previous PR, footer and rebase.

Bottelet:
updated both git-worktrees.md and https.md to describe the cert reissuance and the deep subdomain reasoning. footer stripped.

re the SAN buildup on prune — agreed that's worth addressing, the list only grows. will file an issue for cleanup on worktree removal. the unsecured window between os.Remove and a failed IssueCert is a fair point too, though as you say mkcert makes it low risk in practice.

tests for the nginx wildcard and the domain list builder are on my list — will push those as a follow-up.

PR #310, opened 2026-05-07
feat: built-in Vite dev server worker with host execution and worktree auto-start

Bottelet:
## Summary

Adds a built-in `vite` worker for Laravel that runs `npm run dev` on the host via fnm. Vite auto-starts when worktrees are created, with per-worktree systemd units so multiple instances can run simultaneously.

## Problem

Developers need to manually run `npm run dev` in each worktree for HMR. Forgetting this means stale assets or `ViteManifestNotFoundException`. There's no way to automate it through lerd's existing worker system since all workers execute inside the PHP-FPM container via `podman exec`, and Vite needs to run on the host for direct browser HMR access.

## Solution

**Host execution mode for workers:**
- New `host: true` field on `FrameworkWorker` — generates a systemd unit that runs via `fnm exec` on the host instead of `podman exec` in a container
- Node.js version auto-detected from the project's `.node-version` / `package.json`

**Built-in vite worker:**
- Added to Laravel's framework workers with `check: vite.config.js`
- `mergeBuiltinWorkers()` ensures new built-in workers are picked up even when a store framework definition exists

**Worktree auto-start:**
- `syncWorktree` auto-starts host workers when a worktree is created
- Per-worktree unit naming (`lerd-vite-site-branch`) allows simultaneous Vite instances with auto-incremented ports
- `resolveSiteAndFramework` falls back to parent site for worktree paths so `lerd worker start/stop` works from worktree directories

## Usage

```bash
# Manual start/stop
lerd worker start vite
lerd worker stop vite

# Auto-start on worktree creation (no action needed)
lerd worktree add my-feature
# → Vite Dev Server started for myapp
#   Logs: journalctl --user -u lerd-vite-myapp-my-feature -f

# List workers
lerd worker list
# vite            Vite Dev Server
#                 command: npm run dev
```

Add to `.lerd.yaml` to persist across `lerd start`:
```yaml
workers:
    - horizon
    - vite
```


geodro (changes requested):
The host worker concept is a good fit for vite, and per worktree unit naming with the parent site fallback in resolveSiteAndFramework reads well. CI is currently red on Linux though, the existing worker_schedule_linux_test.go has three calls to writeWorkerUnitFile that need a false 10th arg appended for the new host parameter. Three line fix, macOS passes only because the file is _linux_test.go.

Bigger ask, please move the vite worker definition into the lerd-frameworks repo at https://github.com/geodro/lerd-frameworks (frameworks/laravel/*.yaml) and drop both the hardcoded entry in laravelFramework and the new mergeBuiltinWorkers helper. That repo is where framework workers live, queue, schedule, reverb, horizon and tinker are all defined there, vite should follow suit. The merger introduces a parallel source of truth that we'd rather not entrench, the store yaml needs to be the one place a worker is defined. Happy to review the lerd-frameworks PR alongside this one once it's up.

The bigger behavioural concern is that the new auto start collides with the existing npm run build prompt in lerd worktree add. That flow already asks the user "Build the worktree's frontend assets?" with a Skip option specifically labelled for people who want to run npm run dev themselves. Auto-starting vite dev as a systemd unit makes that Skip path moot and runs npm run build and npm run dev in parallel for every worktree on a Vite project. Either the auto start should suppress the build prompt when it kicks in, or the build prompt should suppress the auto start, but the two paths need to know about each other.

A few smaller behavioural questions worth a thought too. There's no opt out for the auto start, if a project has a vite.config.js syncWorktree spawns a unit whether the developer wants one or not. WorkerStartForSite ends with AddProjectWorker which means auto start also persists vite into the worktree's .lerd.yaml as if the user opted in manually, may be intended but worth being deliberate about it. And when a worktree is pruned the unit isn't stopped or removed so dead lerd-vite-* services accumulate, same shape as the cert SAN buildup we discussed on the previous PR. The hardcoded "22" fallback for node version when detection and global config both fail also reads like a magic number that should at least live as a named constant.

Tests would be welcome, the host unit body and the per worktree unit naming branch in WorkerStartForSite, plus the worktree fallback in resolveSiteAndFramework, are all uncovered. None of those need elaborate fixtures.

Docs gap as well, host: true on workers, the new vite worker, and the auto start on worktree creation aren't mentioned in docs/usage/framework-workers.md, docs/usage/framework-definitions.md, or docs/features/git-worktrees.md.

Same housekeeping as the other PRs, footer and rebase.

Bottelet:
fixed the CI — three missing false args in worker_schedule_linux_test.go, pushed.

moved the vite worker definition out of the built-in laravelFramework and into lerd-frameworks (geodro/lerd-frameworks#6 for laravel 11/12/13). dropped mergeBuiltinWorkers entirely so the store yaml stays the single source of truth.

re the auto-start vs build prompt collision — good catch. the two paths do step on each other right now. i think the cleanest fix is to suppress the build prompt when the framework has a host worker that auto-starts (vite covers what build would have produced). will push that in a follow-up.

the orphan unit cleanup on worktree prune, the opt-out mechanism, and the AddProjectWorker persistence are all fair points. will file issues for those rather than bloat this PR further.

renamed the "22" fallback to a constant. tests and docs coming next.

geodro:
Thanks, the structural cleanup is exactly what I was after, vite in lerd-frameworks, mergeBuiltinWorkers gone, the constant rename, and the host unit tests all look good.

A couple of pieces from the last round still feel like they belong in this PR rather than follow ups though. The build prompt collision and the AddProjectWorker persistence are both behavioural bugs that ship the moment this merges, every Vite worktree will either get a duplicate build or a surprise commit to .lerd.yaml on first add, so I would rather we close them here than land the regression and chase it. Both fixes are small, the prompt one just needs the two paths to know about each other, the persistence one is a flag on WorkerStartForSite for the auto trigger path.

Test coverage and docs gaps from the previous comment are still open too, the per worktree naming branch and the resolveSiteAndFramework worktree fallback aren't pinned by anything, and framework-definitions.md and git-worktrees.md don't yet mention host workers. Happy for the orphan cleanup and opt out to be follow ups as you suggested, just not the two behavioural ones above.

Bottelet:
all addressed in the latest push:

**build prompt collision** — hasHostWorkerAutoStart checks if the framework defines a host worker with a passing check rule. when true, lerd worktree add skips the build prompt and prints a note that vite was auto-started by the watcher. the two paths no longer race.

**AddProjectWorker persistence** — WorkerStartForSite now takes an optional persist flag (variadic bool). the auto-start path in syncWorktree passes false so vite doesn't get written to the worktree's .lerd.yaml. manual lerd worker start still persists as before.

**tests** — added a test for per-worktree unit naming via WorkerStartForSite (asserts lerd-vite-mysite-<dir> pattern). the host unit body and container mode tests were already in the previous push.

**docs** — added host: true to the worker schema in framework-definitions.md, and documented host worker auto-start + build prompt suppression in git-worktrees.md.

the "22" constant was already renamed to defaultNodeVersion in the earlier push.

PR #359, opened 2026-05-15
feat: serve lerd.localhost dashboard over HTTPS by default

erhanurgun:
## Summary
- 443 server block + 301 redirect from port 80, using the existing
  lerd.localhost cert.
- `install` and `start` now issue the cert before writing the vhost.
- Falls back to HTTP-only when the cert is missing so a fresh install
  does not refuse to start.

## Why
Every other .test site is HTTPS by default. The dashboard was not.

## Test plan
- `go test ./internal/nginx/` passes (2 new tests).
- `curl -ksI http://lerd.localhost` returns 301; `https://` returns 200.
- Deleting the cert and re-running `lerd start` produces an HTTP-only
  vhost without error.

## SS

image


geodro:
Hey, thanks a lot for taking the time to put this together. Before we merge I want to push back on the premise though. The dashboard sits on a .localhost domain which is RFC 6761 reserved and always resolves to loopback, so traffic never leaves the machine, and browsers already treat http on localhost as a secure context. Service workers, crypto.subtle and the rest of the platform features work the same as they would over https. The .test sites are a different story because they can be exposed via lerd expose and genuinely cross networks, but lerd.localhost isn't tunnelable. Is there a specific use case driving this on your end, something like LAN access, a remote dev setup or a corporate policy that flags plain http? Happy to reconsider if there's a concrete scenario, just want to make sure we're not adding moving parts for a cosmetic win.

erhanurgun:
Thanks for taking the time to discuss this before declining. To be upfront: this PR is cosmetic, and I just noticed your own comment at [`internal/nginx/manager.go:844-847`](https://github.com/geodro/lerd/blob/42dddd4e227584198cc42759e67923aaf99f5b2d/internal/nginx/manager.go#L844-L847) already documents the same `.localhost` reasoning.

Per KISS and YAGNI, I'll leave the merge or close decision to you. Happy either way. Next time I'll start with an issue or discussion before jumping to a PR.

geodro:
Appreciate the graceful reply. Going to close on the YAGNI side. Thanks for the swing.

PR #409, opened 2026-05-24
Oracle oci8 support

gabriel-sousa99:


geodro:
hey, thanks for putting all this together, but I dont think I can take this PR as it stands. its titled Oracle oci8 support but its really your forks main branch opened against ours, 71 files and around 5,600 additions across 24 commits, and most of it is fork specific or out of scope for a single PR.

the rebrand commit rewrites the README to Lerd Oracle Edition, translates it to Portuguese only, points the installer at gabriel sousa99/lerd, redirects auto update away from this repo, and renames the version scheme to 1.21.2 oracle.N. that all needs to live in your fork, not here. the two top commits, the v1.21.2 release and the MySQL probe / dump bridge / PHP 7.2 fix, are already on main as 225547fe and the v1.21.2 I just cut, so they would only create conflicts.

the other things mixed in here, the PHP 5.6 tier, the debug area redesign, the custom PHP extension manager, the env editors, the install streaming, the artisan auto discovery, the CSRF fix, the install.sh parse fix, the typesense dashboard preset, some of those could be interesting on their own but each would need its own small PR with a short description so I can actually evaluate it, and a few overlap with work already in flight. bundled together like this theres no realistic path to merge any of it.

going to close this one. if you ever want to bring one of the smaller standalone fixes upstream as its own PR Im happy to look at it then.

Issue #428, opened 2026-05-27
Per-service tuning config in the web UI (PHP / MariaDB / Redis)


prabowosd:
There's no surface in the web UI (or a first-class CLI command) for **tuning the runtime config of a service**. The service installs and runs fine, but every meaningful performance/behaviour knob has to be hand-edited in a file under `~/.local/share/lerd/...` (or via `podman exec`), then the container restarted manually. That's undiscoverable from the UI that otherwise manages the whole stack, and easy to lose across a reinstall. Affects PHP, MariaDB/MySQL, Redis — basically every service with a config file.

### What I do today (workarounds)

- **PHP (per version):** edit `~/.local/share/lerd/php/<ver>/98-user.ini` by hand (mounts as `98-lerd-user.ini`), then `podman restart lerd-php8X-fpm` — e.g. `memory_limit`, `opcache.memory_consumption`, `opcache.max_accelerated_files`, `realpath_cache_size`/`ttl` (huge win on the virtiofs mount), `opcache.revalidate_freq`, JIT on/off. None of it reachable from the UI.
- **MariaDB/MySQL:** no managed `my.cnf` override at all. Hit `max_allowed_packet` limits importing large dumps and had to `SET GLOBAL` at runtime; nowhere persistent to set `innodb_buffer_pool_size`, `max_allowed_packet`, `innodb_log_file_size`, etc.
- **Redis:** same — no managed `redis.conf` override for `maxmemory`, `maxmemory-policy`, persistence (`save`/`appendonly`).

### Proposal

A "Config / Tuning" panel per service in the web UI (and a matching `lerd service config <name>` CLI), backed by a managed override file that:

1. **Surfaces the common knobs as fields** (current effective value + sane default) per service type — PHP: `memory_limit`, `opcache.*`, `realpath_cache_size`/`ttl`, `revalidate_freq`, JIT toggle; MariaDB/MySQL: `innodb_buffer_pool_size`, `max_allowed_packet`, `innodb_log_file_size`, `max_connections`; Redis: `maxmemory`, `maxmemory-policy`, `appendonly`/`save`.
2. **Also exposes a raw "advanced" textarea** mapping to the underlying override file (`98-user.ini` / a dropped-in `my.cnf` / `redis.conf` fragment), so power users aren't boxed in.
[...]

PR #429, opened 2026-05-27
feat(services): user-editable runtime tuning override for db services

prabowosd:
Implements the per-service tuning config from #428, scoped to the mysql/mariadb families as an MVP.

### What

A user-editable runtime tuning override per service. Lerd seeds it once with a commented template and **never overwrites it afterward**, so edits survive `lerd service reinstall` and `lerd update`. It's bind-mounted *after* the bundled preset config, so any value set there wins.

```
$ lerd service config mariadb-10-11
# opens ~/.local/share/lerd/service-tuning/mariadb-10-11.conf in $EDITOR,
# then restarts the service so it re-reads the config

$ lerd service config mariadb-10-11 --path   # just print the path (for tooling/UI)
```

The seeded file (`/etc/mysql/conf.d/zz-lerd-user.cnf` inside the container — the `zz-` prefix sorts it after the bundled `lerd.cnf`):

```ini
[mysqld]
# Lerd user tuning for this service. ... never overwritten ...
# max_allowed_packet = 1G
# innodb_buffer_pool_size = 512M
# innodb_log_file_size = 256M
# max_connections = 200
```

### Design notes

- **Not a revival of user `files:`.** This is a *single managed* override per service at a family-determined path — it mirrors the already-blessed never-clobber contracts of the per-version PHP `98-user.ini` and nginx `custom.d`, rather than reintroducing arbitrary user-declared file mounts.
- **One hook, both platforms.** The override is emitted as a `Volume=` line in `GenerateCustomQuadlet`, right after the preset-file loop. The macOS launchd path re-parses the same quadlet `[Container]` section (`Volume=` → `-v` in `containerToPodmanArgs`), so no separate darwin code is needed.
- **Materialize-before-generate.** `MaterializeServiceTuning` runs in `EnsureCustomServiceQuadlet` ahead of generation, so the mounted host path always exists for tunable families. Backward-compatible: existing installs pick up the mount on their next start/restart.
- **DRY-up.** Extracted `launchEditor` and reused it in `php:ini` (identical behaviour).

### Scope / follow-ups (separate PRs)

This PR is backend + CLI for mysql/mariadb only. Planned next:
- Web UI tuning panel (raw advanced editor + curated fields)
- Surface the existing PHP `98-user.ini` in the UI for parity
- redis / postgres families (each needs per-image include handling)

### Testing

`go test ./...` → 2211 passing. New tests cover the family→target mapping (known/inferred/untuned/nil), seed-once + never-overwrite materialization, the quadlet mount emission, and the CLI command (`--path` seeds+prints, rejects untuned families and uninstalled services). Smoke-tested end-to-end against a real `mariadb-10-11` service.


geodro:
Hey @prabowosd, thanks for putting this together.

Pulled this down and ran it against my install. The design holds up well, I registered a custom mariadb 10 11 via a yaml in the services dir and the seed plus path output worked, edited max_allowed_packet, re-ran, the edit was preserved. Setting EDITOR to true with no restart regenerated the quadlet, and the unit ended up with the Volume line mapping the host conf to /etc/mysql/conf.d/zz-lerd-user.cnf ro,z exactly as advertised.

One thing does not work though. Default presets are unreachable. Running lerd service config mysql, or postgres, or redis errors with service "mysql" is not installed even when service list shows it active. The cause is that the new command goes through config.LoadCustomService which only reads ~/.config/lerd/services/<name>.yaml, and default presets never get a yaml on disk, they live in code. The realistic flow that issue 428 calls out, install mariadb via lerd service preset install mariadb and then tune it, hits this immediately. Worth resolving before merge since this is the most common path users take.

Two ways out, you could extend the lookup so default presets resolve to their in code definition the same way service remove already handles built ins, or you could persist a yaml on default preset install so LoadCustomService finds them uniformly. The first is the smaller patch, the second removes a longer standing asymmetry between built ins and custom services more broadly.

Cosmetic, the rejection error for an installed service without an explicit family in its yaml renders family "" in the message, gotenberg in my case. Falling back to InferFamily there would print the inferred name instead of an empty quote.

Also docs/usage/custom-services.md does not yet mention the new command, that should land alongside this.

geodro:
Also, if this is not final yet, would you mind switching it to draft while you address the points above? Makes it easier to track what is ready for a proper review pass versus still in flight.

prabowosd:
Thanks for the thorough pass — and good catch on the default-preset path. Pushed fixes for all three points; switched back to draft as requested.

**Default presets unreachable** — fixed with your first ("smaller patch") suggestion. Added `config.ResolveServiceForTuning`, which falls back to resolving the in-code default-preset definition when there's no YAML on disk (the same built-in-aware lookup `service remove` already uses), and wired the CLI through it. Quadlet regeneration now branches on `IsDefaultPreset` so built-ins regenerate via `EnsureDefaultPresetQuadlet` (which itself resolves to `EnsureCustomServiceQuadlet`, so the mount lands either way). The `lerd service preset install mariadb` → `lerd service config mariadb` flow from #428 works now.

(Scope note: this PR keeps the tunable families to **mysql/mariadb**. `lerd service config redis|postgres` now *resolves* the preset but reports "does not support tuning yet" — the redis/postgres families land in a follow-up PR that adds their `tuningMount` entries + the container-command injection they need.)

**Cosmetic `family ""`** — fixed. When the service has no resolvable family (your gotenberg case, where `InferFamily` also comes back empty), the message now drops the family clause entirely rather than printing empty quotes.

**Docs** — added a "Tuning a service" section to `docs/usage/custom-services.md` covering `lerd service config` (+ `--path` / `--no-restart`) and the never-overwrite contract.

I'll mark it ready for review once you've had a chance to confirm the default-preset flow works on your end.

prabowosd:
For context on where this is heading — I kept this PR deliberately small (backend + CLI, mysql/mariadb), but the rest is already built and will follow as separate, stacked PRs once this lands, so each stays reviewable on its own:

- **Web UI tuning tab** — a "Tuning" tab on the service detail panel (textarea + Save & restart), backed by `GET`/`POST /api/services/{name}/config`. Sets a `tunable` flag on the service response so the tab only shows for tunable services, and reuses `ResolveServiceForTuning` so built-in defaults work there too.
- **PHP `php.ini` editor** — surfaces the existing per-version `98-user.ini` in the System → PHP panel (collapsible editor + Save & restart), via `GET`/`POST /api/php-versions/{version}/config`. Mirrors the `lerd php:ini` CLI command. (Independent of this PR.)
- **redis / postgres families** — adds their `tuningMount` entries plus the container-command injection they need (`redis-server <conf>`, `postgres -c include_dir=…`), since unlike mysql/mariadb their images load no config by default.

All have tests (Go + Vitest) and the UI strings are translated across all 8 locales. Happy to open them in whatever order/grouping you prefer once you're good with the foundation here.

prabowosd:
One more in the same "surface config in the web UI" vein, built as its own independent branch (base `main`, separate PR — not stacked on this one):

**nginx config in the web UI** — two parts:

- **Per-site** — an "nginx" tab on the site detail panel that reads/saves the site's `custom.d/<domain>.conf` override (already included at the end of each server block, but until now it was file-only with no UI or CLI). `GET`/`POST /api/sites/{domain}/nginx`; the domain is validated against registered sites, which also blocks path traversal via the segment. Saving reloads nginx.
- **Global http-level** — a collapsible "http config" editor on the System → nginx panel for http{}-level tuning (gzip, proxy buffers, global `client_max_body_size`, …). Adds `include /etc/nginx/http.d/*.conf;` to the nginx.conf template (loaded last, after lerd's defaults, so user values win), mounts `http.d` into lerd-nginx, and `EnsureHttpD` creates the dir. `GET`/`POST /api/nginx/config`; saving reloads nginx.

Both keep the same seed-template, never-overwrite contract as the service tuning, with Go + Vitest tests and i18n across all 8 locales. The per-site path needs no container change (custom.d is already mounted); the global path adds the one `http.d` mount.

So the full set of follow-ups now: service tuning UI tab, php.ini editor, redis/postgres families, and nginx (per-site + global). Happy to sequence the PRs however works best for you once the foundation here lands.

prabowosd:
Hey @geodro please check

geodro:
Pulled it down and ran through the scenarios on my install. The default preset path is fixed, lerd service config mysql --path resolves cleanly and seeds the template, and the family aware error is back for postgres and redis. The empty family case on gotenberg now prints the message without the empty quotes, that reads much better. Docs section is good too. Nice that this lands the CLI half of 428, the install then tune flow from the issue works without surprises now.

Wired up the override end to end. EDITOR=true with no-restart added the new Volume line to a quadlet that pre-dated the feature, then I set max_allowed_packet to 1G, restarted lerd-mysql, and mysqld picked it up, 67108864 went to 1073741824. Reverted to the seed and restart, back to 64M. Re-running config did not clobber my edit either, so the never-overwrite contract holds in the real flow not just in the unit tests.

One thing worth tightening before merge. ResolveServiceForTuning succeeds on a default preset even when it is not installed, because it falls back to LoadPreset on the in-binary definition. So if someone has explicitly lerd service remove mysql and then runs lerd service config mysql, we will seed the file, regenerate the quadlet via EnsureDefaultPresetQuadlet, and the final RestartUnit will actually start the unit, effectively reinstalling mysql as a side effect of an edit command. Narrow window since the install then tune flow from 428 lands you on an installed service anyway, but the new serviceops.ServiceInstalled from 433 closes it in one check, errors out with run lerd service preset install mysql first if the quadlet is not on disk.

Two smaller things I would fold into the redis and postgres follow-up rather than block on here. service remove --remove-data does not touch the tuning file, the never-overwrite contract is great across reinstalls but on an explicit data-wiping remove users will probably expect the override to go with it. And the supported mysql, mariadb string in the error message is hardcoded, deriving it from tuningMounts keys would keep it honest when the next families land.

Other than the install presence guard this looks ready, happy to take another pass once that lands.

prabowosd:
Quick status update on the broader queue, in case it helps you sequence:

- **#436** (php.ini editor) and **#437** (nginx config UI — per-site + global) are now open. Both are independent of this PR (base `main`), so they're reviewable any time.
- The two remaining follow-ups — **service tuning UI tab** and **redis/postgres families** — are stacked on this branch. I'd rather hold them until this one lands so the diffs stay clean (otherwise they'd carry all of #429's commits along), but they're built, tested, and ready to push the moment you merge.

No rush on this PR — just wanted to surface the dependency so you can prioritise however works best for you.

prabowosd:
Thanks for the thorough run-through — really glad the override roundtrip held up on a pre-feature quadlet too, that was the case I was most worried about.

**Install-presence guard — fixed.** Pushed `9a8471c` (rebased onto current `main` to pick up #433's `serviceops.ServiceInstalled`):

- `service config` now checks `ServiceInstalled(name)` immediately after `ResolveServiceForTuning` and errors with `service "mysql" is not installed — run \`lerd service preset install mysql\` first` when the quadlet isn't on disk. Closes the silent-reinstall-on-edit window you flagged for the removed-then-tuned default-preset path.
- New regression test `TestServiceConfig_RejectsRemovedDefaultPreset` covers exactly that case (default preset resolves, but no quadlet → guard must reject + hint at install). The existing happy-path / untuned-family tests now lay down a fake quadlet in the isolated `QuadletDir` to keep satisfying `ServiceInstalled`.
- Documented the precondition in `docs/usage/custom-services.md`.

**Folded into the redis/postgres follow-up:**

- `service remove --remove-data` cleaning up the tuning file — agreed, the never-overwrite contract should defer to an explicit data wipe. Will land alongside the redis/postgres families branch.
- Deriving the "supported: …" string from `tuningMounts` keys — same branch. That branch is the one that grows the keyset, so it makes sense for the message-honesty fix to land with it rather than churn the string here.

Gate (`go test ./...`, `gofmt -l`, `make build`) green. Ready for another pass whenever you're around.

prabowosd:
Also please check i have already done with my work bro @geodro

PR #436, opened 2026-05-28
feat(ui): edit per-version php.ini from the web UI

prabowosd:
Adds a collapsible **php.ini editor** to the PHP version detail panel, surfacing the existing per-version \`98-user.ini\` file in the web UI. Mirrors the existing \`lerd php:ini\` CLI command — same file, same never-overwrite contract.

Independent of #429 (base \`main\`).

## Changes
- **Backend**: \`GET\`/\`POST /api/php-versions/{version}/config\` — reads/saves \`98-user.ini\` for the given version and restarts the FPM container on save.
- **UI**: \`PhpIniEditor.svelte\` (collapsible textarea + Save & restart) embedded in \`PhpDetail\`, \`getPhpIni\`/\`savePhpIni\` store helpers.
- **i18n**: new keys translated across all 8 locales (de/en/es/fr/id/nl/pt/tr).

## Tests
- \`make build\` — clean

## Notes
Surfaces an already-existing user-config file (\`PHPUserIniFile\`) — no schema/disk-layout change. Existing installs work without any container recreate.


geodro:
Pulled it down and put it on my install, the round trip works, edited memory_limit in the editor, FPM picked up 1024M after the restart, reverted it and it dropped back to 512M. Bundle has the new strings, GET path is clean against URL escape tricks. Few things on the new code though.

The one I would actually block on. On POST, WriteFPMQuadlet is what stitches the 98-user.ini Volume line into the FPM container, and if it fails today the new handler just logs a WARN to stderr and goes ahead and calls RestartUnit anyway. On an install that pre-dates the user-ini feature, where the mount is not in the quadlet yet, that means the freshly written file is orphaned and the container reads nothing, but the API still returns ok true. The user sees a green save with no way to find out the change never made it in. Easiest fix is to return a 500 when the quadlet rewrite errors, or to verify the mount is present after the restart and surface that.

A couple of correctness things to tighten in the same area. The new handler accepts any version that passes the regex shape check and calls EnsureUserIni unconditionally, so a GET on something like /api/php-versions/99.9/config happily seeds a brand new directory under DataDir from nothing, I confirmed it on my box and had a stray 99.9 folder afterward. The CLI php:ini has the same hole but the threat model is different there, this is now reachable from any browser tab, so the new handler should reject versions that are not in the installed set before EnsureUserIni runs. Same area, the POST decodes the body with a bare json.NewDecoder, no MaxBytesReader cap, and the content goes straight to disk via os.WriteFile, so a multi gig POST is decoded into memory and written through. The tinker endpoints in the same file already cap at 64 KiB, easiest to copy that.

UX nit on the save error path. savePhpIni in the new store helper catches everything and returns a boolean, so the editor surfaces only the generic localised string no matter the cause. getPhpIni next to it does the right thing through apiJson and load failures bubble with status and statusText, the asymmetry is jarring. Make savePhpIni throw on non ok the same way load does and let the component render e.message.

Last one is minor. In PhpIniEditor the saved flag is set true with a 2500ms timer to clear itself, but the timer id is not captured and the effect on version change resets open and loaded and content and error and does not touch saved. Switch versions inside 2.5 seconds of a save and the new version shows Saved until the old timer fires. Capture the id, clear it in the cleanup, add saved to the reset list.

Mergeable once the silent quadlet-write failure is addressed, the rest could land here or as immediate follow ups, your call.

prabowosd:
Thanks for the careful pass — every one of these turned out to be a real foot-gun once I looked. All five addressed in `99535ae`:

**Blocker — silent quadlet-write failure.** `WriteFPMQuadlet` is now a 500. The orphaned-write-with-green-save path you described is closed; the editor surfaces the server's reason via `e.message` (see UX fix below).

**Phantom version directory.** Handler now rejects with 404 if `version` isn't in `phpPkg.ListInstalled()` before `EnsureUserIni` runs. GET `/api/php-versions/99.9/config` no longer seeds anything.

**Unbounded POST body.** Capped via `http.MaxBytesReader(w, r.Body, 64<<10)`, matching the tinker endpoints in the same file.

**Save error symmetry.** `savePhpIni` now throws on non-ok with the server body as the message (mirrors `getPhpIni`'s `apiJson` behaviour). The editor renders `e.message` and falls back to the localised generic string only when the error has no message.

**Saved-flag timer.** Captured the `setTimeout` id, clear it in the `clearSavedTimer()` helper, and the version-change effect now resets `saved` + clears the pending timer. Switching versions inside 2.5s of a save no longer leaks a stale "Saved" badge to the new version.

Tests cover the changed contract: vitest now asserts `savePhpIni` resolves void on ok and rejects with the server body on non-ok. Gate (`go test ./...`, `gofmt`, `npm run check`, `npm test`) green across the board.

PR #437, opened 2026-05-28
feat(nginx): edit nginx config (per-site + global http-level) from the web UI

prabowosd:
Adds two nginx config editors to the web UI, both keeping the same seed-template / never-overwrite contract as the service tuning work in #429.

Independent of #429 (base \`main\`).

## Part A — Per-site override (\`custom.d/<domain>.conf\`)
\`custom.d\` is already included at the end of each site's server block, but until now it was file-only with no UI or CLI. This adds an **"nginx" tab** to the site detail panel that reads/saves the override.

- Backend: \`GET\`/\`POST /api/sites/{domain}/nginx\` — domain validated via \`FindSiteByDomain\`, which also blocks path traversal through the \`{domain}\` segment. Saving reloads nginx.
- UI: \`SiteNginxTab.svelte\` + \`getSiteNginx\`/\`saveSiteNginx\` store helpers.
- **No container change needed** — \`custom.d\` is already mounted. Works on existing installs as-is.

## Part B — Global http-level config (\`http.d/zz-lerd-user.conf\`)
For http{}-level tuning (gzip, proxy buffers, global \`client_max_body_size\`, …). Adds a collapsible **"http config"** editor on the System → nginx panel.

- Template: \`include /etc/nginx/http.d/*.conf;\` appended at the end of the http{} block, so user values win over lerd defaults.
- \`lerd-nginx.container\`: mount \`http.d\`; \`nginx.EnsureHttpD\` creates the dir.
- Backend: \`GET\`/\`POST /api/nginx/config\`. Saving reloads nginx.
- UI: \`NginxConfEditor.svelte\` + \`nginx\` store, embedded in \`NginxDetail\`.
- **Needs the nginx container recreated** (one new mount). \`lerd-heal\` was updated locally to handle this for existing installs; happy to fold a heal/reconcile hook into this PR if you'd prefer.

## i18n
All new keys translated across all 8 locales (de/en/es/fr/id/nl/pt/tr).

## Tests
- \`make build\` — clean

## Notes
UI-only by design (no new CLI command). Avoids a \`launchEditor\` name clash with the CLI introduced by #429 — happy to add a CLI surface in a follow-up once the editor helper lands on main.


geodro:
Resolved the trivial sites.ts conflict locally and put it on my install, both editor surfaces are wired through and the bundle has the new strings.

Per site custom.d round trip is clean. GET returns the seed without writing, POST lands the file, nginx reload picks it up, I added X-PR-437 hello on astrolov.test and saw it on the HTTPS response immediately, removed the file and it's gone. Works on existing installs because custom.d has been mounted into lerd nginx forever, no container change needed.

The global http.d editor is where the silent fail you flagged in the description actually bit. On my install the nginx quadlet still has conf.d and custom.d Volume lines but no http.d mount, and the rendered nginx.conf in the data dir has no include http.d line either. The new template hits both of those but only when lerd start or lerd update re renders, so a user who swaps the binary in place and goes straight to the editor will POST, get ok true, and write a file that nginx never sees. Easiest fix is to call EnsureNginxConfig plus the quadlet rewrite path inside the handler on first POST so the editor heals its own preconditions, similar to how the php ini handler in 436 calls WriteFPMQuadlet before restart. That, and Part A is already mergeable.

Two UI requests, both soft, I can do them on top after merge if you would rather keep this PR focused on the backend wiring. First, ideally style the two editors after the env tab from 434, same textarea framing, save button on the right, save row layout, font sizing, the current ones are close but a notch off and inconsistent across the two surfaces. Second, mirror the env tab's save flow as well, modal confirm on save with a checkbox for backup, plus a revert feature that lists prior backups and lets you restore one, the never overwrite contract for these files makes accidental edits more likely than for env so the safety net pays off more. Neither blocks.

Branch is now conflicting with main since 429 went in, easy to clear with the Update branch button on the PR.

prabowosd:
Nice catch on the silent failure — that one was clearly the editor's job to heal, not the user's.

**Blocker — silent quadlet/include miss.** Fixed in \`0c2d11d\` (rebased onto current \`main\` so the \`sites.ts\` conflict is resolved here too, no more local fixup needed):

\`handleNginxConfig\` now heals its own preconditions on POST, mirroring the \`WriteFPMQuadlet\` pattern from #436:

- \`nginx.EnsureNginxConfig()\` rerenders \`nginx.conf\` from the bundled template (now carrying \`include /etc/nginx/http.d/*.conf;\`) and creates the \`http.d\` dir.
- \`nginx.RewriteNginxQuadlet()\` (new helper) rewrites \`lerd-nginx.container\` from the bundled template (now carrying the \`http.d\` \`Volume=\`) via \`WriteQuadletDiff\` and reports whether the on-disk quadlet actually changed.
- If the quadlet changed → \`DaemonReload\` + \`RestartUnit("lerd-nginx")\`. \`nginx -s reload\` won't pick up a new \`Volume=\` mount, so a restart is mandatory the first time.
- If the quadlet was already current (every subsequent save) → plain \`nginx.Reload()\`.

So on an install predating http.d, the first POST self-heals (with the one container restart that's actually needed); from then on it's reload-only. The orphan-write path is gone.

**Also folded in:** capped the POST body at 64 KiB via \`MaxBytesReader\` to match what you flagged on #436 — same exposure on the same shape of handler.

**Regression guards added:**
- \`TestNginxQuadletMountsHttpD\` — bundled template carries the \`http.d\` \`Volume=\`.
- \`TestEnsureNginxConfig_writesForwardedAndCustomD\` extended to also assert \`http.d\` dir + the \`include\` directive in the rendered \`nginx.conf\`. Both preconditions the heal relies on staying no-op once it has run.

**UI requests** (style after env tab, modal confirm with backup + revert): both definitely worth doing, but I agree they're a meaningful chunk of UX work and I'd rather not bundle them with the precondition fix — happy to land them as a focused follow-up PR once this merges, so the env tab pattern (\`SaveEnvResult\`, \`SiteEnvBackup\`, backup checkbox, restore listing) can be lifted across both editors cleanly. Let me know if you'd prefer it inlined here instead and I'll fold.

Gate (\`go test ./...\` 2346 pass, \`gofmt\`, \`npm run check\` 0/0, \`npm test\` 367 pass) green.

geodro:
Heal flow on the global editor is exactly right, my lerd-nginx quadlet had no http.d mount before this and nginx.conf had no include line, one POST to api/nginx/config now writes the override, rewrites the quadlet via RewriteNginxQuadlet, rerenders nginx.conf via EnsureNginxConfig, daemon reloads, restarts lerd-nginx because a new Volume needs container restart, and X-PR-437-R2 healed came back on https in the response. The two regression tests anchor both preconditions so this won't drift. Body cap on the global handler also works, 100 KB POST returns 400 invalid request body as expected. Clean fix.

One leftover on the per-site handler though. handleSiteNginx in the same file still does json.NewDecoder on r.Body with no MaxBytesReader, and the cap is the same hardening you just added one screen down. I sent a 100 KB body to api/sites/astrolov.test/nginx, the body was decoded, 100,000 bytes were written to custom.d/astrolov.test.conf, and only then nginx rejected the garbage with too long parameter and returned 500 saved but nginx reload failed. The disk write already happened. One line copy of the 64 KiB cap from handleNginxConfig fixes it, same shape.

Otherwise this is ready, the env tab style polish is on me for after merge per my earlier note.

prabowosd:
Good catch — fixed in \`d45bc2c\`. Capped the per-site handler at the same 64 KiB with \`http.MaxBytesReader\`, matching \`handleNginxConfig\` one screen down and the php.ini / tinker endpoints. A 100 KB POST now fails fast at decode with 400 instead of landing on disk first and only being rejected by nginx at reload time.

prabowosd:
Also confirming on this one: env-tab polish across the two nginx editors is all yours per your note above. I won't open a follow-up PR for it.

PR #438, opened 2026-05-28
feat(ui): service tuning config panel

prabowosd:
Builds on the CLI tuning workflow from #429: adds a **Tuning tab** to the service detail panel so the same per-service runtime override (mysql/mariadb's \`zz-lerd-user.cnf\`, etc.) can be edited from the dashboard. Same seed-once, never-overwrite contract; saving restarts the service so the new values take effect.

## Backend
- New \`Tunable\` flag on \`ServiceResponse\` so the UI only renders the tab for services that expose a tuning mount. Derived from \`config.ServiceTuningMount\` against the resolved service.
- \`GET\`/\`POST /api/services/{name}/config\` — same materialise / write / regen-quadlet / restart pipeline as the CLI. Uses \`config.ResolveServiceForTuning\` so **built-in default presets** (not just custom-YAML services) are editable. Quadlet regeneration branches on \`IsDefaultPreset\` to land on \`EnsureDefaultPresetQuadlet\` for built-ins.

## UI
- \`ServiceTuningTab.svelte\` — collapsible textarea + Save & restart, mirroring the php.ini editor's shape. Embedded in \`ServiceDetail\` and only mounted when the service is \`tunable\`.
- \`getServiceConfig\`/\`saveServiceConfig\` store helpers.
- i18n: all new keys translated across all 8 locales (de/en/es/fr/id/nl/pt/tr).

## Tests
- \`make build\` — clean

## Notes
The install-presence guard merged in #429 (\`serviceops.ServiceInstalled\`) applies on the CLI side. The handler here resolves through \`ResolveServiceForTuning\` so removed default presets still produce the same "does not support tuning" / "not installed" surface area; happy to add an explicit \`ServiceInstalled\` check in the handler too if you'd prefer symmetry with the CLI.


geodro:
Pulled it down and put it on my install, the Tunable flag wires correctly so only mysql shows the Tuning tab among my services, redis and postgres correctly return does not support tuning at 400, mariadb returns not installed at 404, and the happy path POST on mysql edits max_allowed_packet to 1G, restart picks it up, restoring the seed reverts to 64M. The end to end shape is right.

A few correctness things to nail down before this merges though, and most of them are direct carryovers from the round one feedback on 436 that didn't propagate forward.

To answer your direct question in the description, yes please add the ServiceInstalled guard symmetrically. Same threat model as 429, removed default preset plus a click on the Tuning tab silently reinstalls the service via materialise plus quadlet regen plus restart, the only thing that changes is the surface. The one liner from 9a8471c on the CLI side ports over verbatim.

The silent quadlet regen failure is the exact same shape that 436 round one had with WriteFPMQuadlet, you fixed it there in 99535ae by converting the WARN to an http.Error 500. The new config branch still does fmt.Fprintf to stderr and proceeds to RestartUnit on both EnsureDefaultPresetQuadlet and EnsureCustomServiceQuadlet failures. On an install pre dating the tuning mount, a regen failure orphans the freshly written file, the restart succeeds with no mount, and the API returns ok true. Same patch as 436 lands here.

POST body has no MaxBytesReader either, I sent 100 KB of garbage and watched 100,000 bytes land on disk with HTTP 200, mysql restarted with it. 64 KiB cap matching the tinker endpoints, same as 436 round two.

Knock on from the install presence concern, buildServiceResponseWithPortList sets Tunable true whenever ResolveServiceForTuning succeeds, which is also true for removed default presets, so the Tuning tab actually renders in the UI for a service that's not installed and clicking it triggers the silent reinstall path. Add a ServiceInstalled gate around the Tunable assignment too.

UX wise, saveServiceConfig in the store catches everything and returns a boolean, so the editor surfaces the generic localised string no matter what failed, same asymmetry savePhpIni had in 436 round one. Throw on non ok with the server body, same as savePhpIni now does. And the saved indicator timeout is not captured or cleared, switching services within 2.5s of a save leaks the orphan timer into the new service context, same bug PhpIniEditor had in round one, same capture plus clearSavedTimer pattern fixes it.

One structural note. The materialise plus write plus regen quadlet plus restart sequence is now inlined in both the CLI service_config.go and this HTTP handler, with the install presence guard sitting on only the CLI side. xdebugops.Apply is the established pattern in the codebase for sharing exactly this kind of operation between CLI and UI. A serviceops.SaveTuningOverride helper that both call would mean the guard, the MaxBytesReader, and the regen failure fix all land once instead of being requested twice, and the next surface that needs to save tuning, MCP for instance, doesn't grow a third copy.

UI style ask same as 437, this is the fourth textarea plus path plus save in a row and none of them match the env tab's UX yet, ideally modal confirm with a backup checkbox plus a revert from prior backups view. Soft, I can do it on top after merge if you'd rather keep this PR focused on the backend.

Three correctness blockers above, the rest could land here or in a tight follow up, your call.

prabowosd:
Thanks for the thorough pass — fair call on all three correctness items being direct carryovers from #436 round 1. All addressed plus the structural refactor in \`323f6c5\`:

**Structural — \`serviceops.SaveTuningOverride\` (xdebugops.Apply-shaped).** New \`internal/serviceops/tuning.go\` houses the shared pipeline so the CLI, this handler, and any future surface (MCP) can't drift:

- \`SaveTuningOverride(name, content)\` is the single full path: install-guard → resolve → family check → materialise → write → \`EnsureTuningQuadlet\` → \`RestartUnit\`. Returns typed sentinels \`ErrTuningServiceNotInstalled\` and \`ErrTuningFamilyUnsupported\` so the HTTP handler maps them to 404/400 via \`errors.Is\` instead of stringly-typing the error surface.
- \`EnsureTuningQuadlet(name, svc)\` is the regen-only tail (branches on \`IsDefaultPreset\` → \`EnsureDefaultPresetQuadlet\` vs \`EnsureCustomServiceQuadlet\`). Shared between \`SaveTuningOverride\` and the CLI command, which has \`--no-restart\` and so can't use the full pipeline. **Regen failures propagate** as errors, never as WARN-and-continue — that was the same shape \`WriteFPMQuadlet\` had in #436 round 1.
- The CLI \`lerd service config\` now calls \`EnsureTuningQuadlet\` instead of duplicating the regen branch *and* instead of swallowing the failure. Guard + MaxBytesReader + regen-fix all land once now, and a third surface picks them up for free.

**Three correctness blockers — fixed:**

- \`ServiceInstalled\` guard ported verbatim from \`9a8471c\` to the HTTP handler (and lives inside \`SaveTuningOverride\` so any caller inherits it). 404 on removed default presets, no silent reinstall via the regen+restart path.
- Silent quadlet regen failure: now a 500. Both branches (\`EnsureDefaultPresetQuadlet\` and \`EnsureCustomServiceQuadlet\`) go through the same propagating helper, no more \`fmt.Fprintf\` to stderr.
- 64 KiB \`MaxBytesReader\` cap on the POST body, matching tinker / php.ini / nginx.

**Tunable-on-removed-preset knock-on — fixed:** \`buildServiceResponseWithPortList\` now gates the \`Tunable\` assignment on \`ServiceInstalled\` too, so the Tuning tab no longer renders for a service that's not installed. The handler guard is the belt; this is the suspenders.

**UX symmetries with #436 round 2 — fixed:**

- \`saveServiceConfig\` now throws on non-ok with the server body as the message, mirroring \`savePhpIni\` / \`getServiceConfig\`. The editor surfaces \`e.message\` (e.g. *"service \`mysql\` is not installed — run \`lerd service preset install mysql\` first"*) instead of the generic localised string. Vitest covers both the resolves-void-on-ok and rejects-with-body-on-non-ok shapes.
- \`ServiceTuningTab\` captures the \`saved\` \`setTimeout\` id, clears it via \`clearSavedTimer()\` on both the cleanup and the version-change \`$effect\`. Switching services inside the 2.5s window no longer leaks a stale "Saved" badge.

**Tests:**
- \`TestSaveTuningOverride_NotInstalled\` — sentinel + install-hint anchored.
- \`TestSaveTuningOverride_FamilyUnsupported\` — short-circuits before disk writes, sentinel anchored.
- Existing handler / CLI tests still pass through the helper.
- Vitest: 366 pass incl. the new throws-on-non-ok shape and \`locales.test.ts\` key-drift.

Gate green across all four surfaces (\`go test ./...\` 2347 pass, \`gofmt -l\`, \`npm run check\` 0/0, \`npm test\` 366 pass, \`make build\`).

**UI style ask** — same as my reply on #437, I'd rather land the env-tab pattern (modal confirm + backup checkbox + revert listing) as one focused follow-up across all three editors, so they end up actually identical instead of each one being almost-but-not-quite. Happy to inline here if you'd prefer.

prabowosd:
Quick heads-up across the queue: from my side #437, #438 and #439 are all settled now, no further changes planned. Whenever you've got a window to take another look they're ready.

And confirming: env-tab polish across the four editors is all yours per your earlier note. I won't open a follow-up PR for it. Thanks again for the careful reviews on this stack.

PR #439, opened 2026-05-28
feat(services): tuning override for the redis family

prabowosd:
Extends the service tuning workflow merged in #429 to **redis** and **postgres**. Unlike mysql/mariadb (whose images auto-include \`/etc/mysql/conf.d\`), these images load no config by default, so the family declares a container command that points the image at the override:

- \`redis\` → \`redis-server /etc/redis/lerd-user.conf\`
- \`postgres\` → \`postgres -c include_dir=/etc/postgresql/conf.d\`

The quadlet generator emits that command only when the service has no \`Exec\` of its own, so a user-declared \`Exec\` always wins. Same seed-once, never-overwrite contract as mysql/mariadb.

## Changes
- \`config/service_tuning.go\`: \`tuningMounts\` gains \`redis\` (\`/etc/redis/lerd-user.conf\`, with \`Command\`) and \`postgres\` (\`/etc/postgresql/conf.d/zz-lerd-user.conf\`, with \`Command\`) entries; commented templates for both.
- \`config/service_tuning.go\`: new \`TuningFamilies()\` helper returning sorted \`tuningMounts\` keys — used by the CLI's "supported: …" hint instead of a hardcoded string, so the message stays honest when new families land.
- \`podman/quadlet_generate.go\`: emits the family's tuning \`Command\` when the service has no \`Exec\`.
- \`cli/service_config.go\`: \`supported: …\` derived from \`TuningFamilies()\`.
- \`serviceops/remove.go\`: \`RemoveService\` deletes the tuning override file when \`RemoveOptions.RemoveData\` is true. The never-overwrite contract still holds across regular \`service remove\` + \`reinstall\`, but an explicit data wipe takes the override with it — silently applying a stale override to the next install would be surprising. Tolerant of \`ENOENT\`.

## Tests
- \`TestTuningFamilies_SortedAndComplete\` — anchors the hint shape so accidental formatting drift is caught.
- \`TestRemoveService_WithData_DeletesTuningOverride\` — covers the new cleanup.
- \`TestRemoveService_WithoutData_KeepsTuningOverride\` — companion guarding the never-overwrite contract on plain \`service remove\`.
- Existing redis/postgres quadlet-generation tests cover the \`Command\` injection (\`quadlet_embed_test.go\`).
- \`go test ./...\` (2359 pass), \`gofmt -l\`, \`npm run check\` (0/0), \`npm test\` (363 pass), \`make build\` — all green.

## Notes
Runtime behaviour (redis reading the conf, postgres include_dir resolution) was verified separately on a live stack while the branch was at \`local-integration\`; generation + command selection are covered by unit tests.

Closes the two follow-ups you flagged on #429 (\`--remove-data\` cleanup + supported-string honesty); both fit naturally with this branch since it's the one that grows the family set.


geodro:
The supported string derivation and the remove-data cleanup are clean, both follow ups from 429 land exactly as expected, and the redis side works end to end on my install, triggered quadlet regen via service config redis, set maxmemory 256mb plus allkeys-lru in the override, restart picked them up and redis-cli came back 268435456 with the right policy. The user-declared Exec wins branch is anchored by a test too, that's a critical guard.

Postgres though, the command shape doesn't work. On my postgis 18-3.6-alpine the new Exec=postgres -c include_dir=/etc/postgresql/conf.d is rejected immediately with FATAL unrecognized configuration parameter "include_dir" and the container goes into start-limit-hit. The reason is that include_dir is a postgresql.conf directive, not a runtime parameter, so -c can't set it, postgres parses include_dir during config file load before runtime parameter setting happens. My postgres needed a manual quadlet edit to come back. The existing TestGenerateCustomQuadlet_PostgresTuningInjectsIncludeDir only asserts the string appears in the generated quadlet, never that postgres actually accepts the flag, which is why this slipped. The note in the PR description about verifying runtime behaviour on a local-integration stack must have hit a different image or postgres version, on lerd's default it's a hard fail.

To actually get postgres to read an external conf.d the include directive has to land inside postgresql.conf before postgres starts, the -c flag can't do that. Options I can see, either ship an entrypoint wrapper that appends include_dir to postgresql.conf and execs postgres, or split postgres out of this PR and land it later with a working mechanism. The redis half is good to merge either way.

prabowosd:
You're right, that's a hard fail. I had verified the redis half end-to-end but the postgres "verification" was clearly on a different image — sorry for the noise.

Pushed df4002a that drops postgres from this PR entirely. tuningMounts now carries only mysql/mariadb/redis, the postgresTuningTemplate constant is gone, and the test that previously slipped this through (TestGenerateCustomQuadlet_PostgresTuningInjectsIncludeDir was string-asserting the broken Exec line, no real-image runtime check) is removed too. Left an inline note where the postgres entry used to live, explaining the config-load-order issue so a future contributor doesn't put it back the same way.

PR title is now scoped to redis only. The --remove-data tuning-file cleanup and the TuningFamilies() helper for the derived "supported: …" hint both still ride along — they aren't postgres-specific.

I'll open the postgres follow-up with the entrypoint-wrapper approach after this lands, with proper runtime verification on the postgis variant lerd ships by default.

Gate green: go test ./... 2357 pass, gofmt clean, npm run check 0/0, npm test 363 pass.

Issue #457, opened 2026-05-31
🐛 Bug Report: "File not found" when linking projects from external storage (exFAT/APFS) on macOS


nawazsarwar:
When attempting to add or link a project directory located on external storage (such as an external SSD or USB Flash Drive) using Lerd on macOS, the site is added successfully but fails to serve. Instead, accessing the site results in a "File not found." error.

This issue persists regardless of the external drive's file system (tested on both exFAT and APFS) and occurs even though the file permissions (rwx) match those of working projects on the local internal disk.

Steps to Reproduce
Connect an external SSD or Pen Drive formatted in APFS or exFAT to a Mac.

Open the Terminal and navigate to the project directory on the external drive (e.g., cd /Volumes/ExternalDrive/my-project).

Run the command:

Bash
lerd link
(Alternatively: Attempt to add the directory via the Lerd GUI/browser).

The project is successfully registered/added to Lerd.

Navigate to the local development URL in the browser (e.g., http://my-project.test).

Expected Behavior
The site should load and serve files normally, just like projects hosted on the internal Mac storage.

Actual Behavior
The browser displays a plain text error:

File not found.

Environment / Technical Details
OS: macOS

Lerd Version: 1.22.1

External Storage File Systems Tested: exFAT, APFS

Permissions: Verified as rwx (read, write, execute), identical to working local projects.

geodro:
Thanks for the report. I have traced it and the good news is it is nothing to do with your permissions, those are a red herring. On macOS lerd runs everything inside a Podman virtual machine, and that VM only gets your home directory shared into it. Anything on an external drive lives under /Volumes, which is outside home and never gets shared into the VM, so the files simply are not visible from inside the container. The site registers fine because that is just bookkeeping on the host, but when a request comes in the PHP process inside the VM goes looking for your project files, cannot find them, and returns that plain File not found. That is also why it reproduces on both exFAT and APFS, the filesystem is not the factor, the VM share is.

We will put together a fix that shares /Volumes into the Podman machine so external drives are visible inside it.

In the meantime the reliable workaround is to keep your project somewhere under your home directory, for example move or clone it into ~/Sites or similar and link it from there, and it will serve normally. I will update this issue once the external drive support lands.

nawazsarwar:
Thanks alot for the reply, I have tried creating a fix and have created a pull request for your consideration
Thanks 🫡

PR #458, opened 2026-05-31
fix(macos): mount system directories to Podman machine during initial…

### LINKED ISSUE #452: macOS: service config/tuning changes don't apply until launchd plist is reloaded (Restart only kickstarts the cached job)
## Summary

On macOS, changing a running container service's config — most visibly a tuning override via `lerd service config <name>` — regenerates the `.container` quadlet **and** rewrites the launchd `.plist` on disk correctly, but the change never reaches the running container. The container keeps running with its old create-time args (old volumes, old command) until the plist is fully reloaded with `launchctl bootout` + `bootstrap`.

This affects **every tunable family** (mysql / mariadb / redis, and any service whose quadlet changes while running), because the gap is in the shared macOS service-lifecycle code, not in any one family.

## Root cause

In `internal/services/launchd_darwin.go`:

- `WriteContainerUnit` writes the new `.plist` to disk but does **not** reload it into launchd (no `bootout`/`bootstrap`).
- `DaemonReload()` is a **no-op** on macOS (`// launchd picks up plist changes on bootstrap`).
- `darwinServiceManager.Restart` runs `launchctl kickstart -k <label>`, which **re-runs the cached job definition** launchd loaded at the last bootstrap — i.e. the *old* `ProgramArguments` — not the freshly-written on-disk plist.

So the apply path `EnsureTuningQuadlet` → write quadlet + write plist → `DaemonReload` (no-op) → `Restart` (kickstart) leaves launchd executing stale `podman run` args. `podman inspect <name> --format '{{.Config.CreateCommand}}'` shows the old args even though the on-disk plist already has the new ones.

On Linux this works because `DaemonReload` really runs `systemctl --user daemon-reload` (`DaemonReloadIfNeeded`) and the systemd restart recreates the container from the reloaded unit.

## Reproduction (macOS, arm64)

1. Have a running tunable service, e.g. `redis` or `postgres`.
2. Set an active directive in its tuning override and apply it via `lerd service config <name>` (or the `service_config` MCP write).
3. `cat ~/Library/LaunchAgents/lerd-<name>.plist` → the new args (e.g. the tuning `-v` mount / command) are present.
4. `podman inspect lerd-<name> --format '{{.Config.CreateCommand}}'` → still the **old** args; the tuning override is not in effect.
5. `launchctl bootout gui/$(id -u)/com.lerd.lerd-<name>` + `launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/lerd-<name>.plist` + start → now the container runs with the new args and the override applies.

(Found while validating per-service postgres tuning end-to-end; the postgres mechanism itself works once the plist is reloaded.)

## Suggested fix

On macOS, when a container unit's plist content actually changed, the apply/restart path should reload the job (`bootout` + `bootstrap`) instead of relying on `kickstart -k`. Options:

- Have `WriteContainerUnit` detect a content change and reload (bootout+bootstrap) so the next `Restart`/start picks it up; or
- Make `darwinServiceManager.Restart` reload-then-start for container units rather than `kickstart -k`; or
- Track a "plist changed" flag (mirroring Linux's `DaemonReloadIfNeeded`) and bootout+bootstrap when set.

Care needed around the `RunAtLoad=false` + `--restart=always` interplay so the reload reliably recreates the container without leaving an orphaned podman container.

nawazsarwar:
…ization on macOS (#457) Closes #452

### DIFF
--- a/internal/cli/startstop_darwin.go
+++ b/internal/cli/startstop_darwin.go
@@ -3,6 +3,8 @@
 package cli
 
 import (
+	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


geodro:
Thanks for jumping on this, really appreciate you turning a fix around so fast. The approach looks right to me, replacing the default home mount with the standard /Users, /private, /var/folders set and adding /Volumes is a clean way to get external drives visible inside the machine.

A couple of things I am still wrapping my head around, more for my own understanding than anything. The new flags sit in the first init branch, the len(machines) == 0 path, so as far as I can tell they only take effect on a brand new machine. That made me wonder about people who already have a podman machine, possibly yourself, since I think they would keep the old mounts until the machine is recreated. If that is the case it might be worth also nudging existing users to recreate, or detecting it for them, so the fix reaches them too. Curious whether you ran into that at all on your side.

The other thing is just me double checking my own mental model. On macOS /var is a symlink to /private/var, so I think /var/folders already lives under /private and the two mounts overlap. If init went through cleanly for you with both, that is good enough for me, I only mention it in case virtiofs grumbled about a nested mount.

nawazsarwar:
Made certain additional changes considering point(s) in your last comment and compiled it succesfully, uninstalled and installed the local compiled copy seems to work good, though noticed an odd behaviour sometimes, after adding a site and accessing the same gives "File not found", but If we do lerd stop && lerd start, it happens to work. I am not certain if this has something to do with my recent commit. 
Please enlighten

geodro:
Thanks for the iteration. I exercised the JSON-edit path locally against a real `podman-machine-default.json` that didn't have `/Volumes`. `checkMissingMounts` correctly flagged the missing mount, `ensurePodmanMachineMounts` appended the right block (tag matched the test value), the backup file landed, and it was idempotent on re-run. I also walked the quadlet side: `InjectExtraVolumes(ExtraVolumePaths())` already wires non-home site paths into lerd-nginx's quadlet automatically, so once the VM has `/Volumes` (your fix) the container will pick it up end-to-end. Nice.

**One thing you'll need to fix to land this:** CI is red on `gofmt -l .`. Line 131 of `internal/cli/startstop_darwin.go` has trailing whitespace after `}` on the blank line. `gofmt -w internal/cli/startstop_darwin.go` and a force-push will clear both "Build & Test" jobs.

On the "File not found" symptom: I don't think it's coming from your code path here, but my best guess is it isn't quite as random as it looks. Adding a site goes through `siteops.LinkSite`, which writes a new vhost file to `~/.local/share/lerd/nginx/conf.d/` and then calls `nginx.Reload()` (`podman exec lerd-nginx nginx -s reload`). That conf.d directory is bind-mounted into the container via virtiofs. With Podman's applehv backend, virtiofs has cases where a newly created file on the host isn't visible inside the VM for a brief window. nginx reloads, doesn't see the new vhost (or sees it but FPM hasn't picked up the new socket yet for `.php` sites), serves the default vhost's "File not found", and your `lerd stop && lerd start` shakes the cache loose. It would also explain why it's intermittent. If you can reproduce, the smoking gun would be: after the failure, `podman exec lerd-nginx ls /etc/nginx/conf.d`. If your new domain's `.conf` file is missing inside the container while present on the host, that's virtiofs. If it's present but nginx is serving the default vhost, it's a reload timing issue.

I have a small follow-up planned to land after this merges (extract a `virtiofsTag` helper so the test actually exercises production, atomic write for the JSON edit, minor comment cleanup). Nothing you need to do for that. Thanks again.

geodro:
Update: I take back what I said about the launchd `Restart` rewrite being unrelated to "File not found". I reproduced the symptom on current main (your fix not applied) by linking a site outside `$HOME`. The launchd plist on disk gets rewritten correctly by `RewriteFPMQuadlets`, with the new `Volume=` entry. But `services.Mgr.Restart`'s old `launchctl kickstart -k` only restarts the process; it doesn't reload the plist file, so launchd hands the same cached arguments to podman, and the new bind mount never lands in the running container. The FPM stat for the new path fails inside the container, and nginx returns the upstream's "File not found". `lerd stop && lerd start` works because Stop bootouts the unit, which forces the next bootstrap to actually re-read the plist from disk.

So your `bootout + bootstrap` rewrite is the right fix and it does belong in this PR. Sorry for sending you down the virtiofs path. The `gofmt -w` is still the only thing blocking merge from my side.

(Separately: linking a site under `/tmp` will still fail after this PR because `/tmp` isn't mounted in the Podman VM either, same class of issue as `/Volumes`. I'll open a follow-up for that; nothing for you to fix here.)

geodro:
One small request: could you add `Closes #452` to the PR body? That issue (filed by sdp) describes exactly the launchd `kickstart -k` cached-plist bug your `Restart` rewrite fixes, and the keyword will auto-close it on merge.

nawazsarwar:
> One small request: could you add `Closes #452` to the PR body? That issue (filed by sdp) describes exactly the launchd `kickstart -k` cached-plist bug your `Restart` rewrite fixes, and the keyword will auto-close it on merge.

Added Closes #452 as requested

PR #469, opened 2026-06-02
Personal .env.lerd_override for per-project overrides

geodro:
Lets each developer override the values lerd writes into a project's .env without hand-editing the file after every lerd env run, and optionally point a project at their own service instead of lerd's container.

The override lives in a personal .env.lerd_override at the project root. It is plain dotenv syntax and never committed, since lerd keeps it in .gitignore. Every KEY=VALUE in it is layered on last when lerd env runs, so it wins over lerd's defaults and computed values like DB_DATABASE and APP_URL, and values are written verbatim so quotes are preserved. The one reserved key, LERD_EXTERNAL_SERVICES with a comma separated list, marks services lerd should write connection vars for but not start or provision, which is how a project keeps using your own local postgres while lerd manages everything else.

A new lerd env:override command scaffolds the file from a commented template, seeds KEY=VALUE arguments, and makes sure it is gitignored. The same flow is exposed to AI assistants through a new env_override MCP tool, and lerd env:check now skips the file since it is a partial overlay rather than a full env file. It stays editable in the dashboard Env tab because it is a file you manage by hand.

While wiring the MCP tool I noticed the injected skill and guidelines had drifted: nine tools (service_config, service_check_updates, workers_health, workers_heal, framework_search, framework_install, dns_diagnose, analyze_queries, site_nginx) and the commands family were registered but undocumented. They are all documented now, and a guard fails the build if any registered MCP tool is ever missing from the skill or guidelines again.


danielhe4rt:
Thanks for accepting the suggestion, mate!

PR #471, opened 2026-06-03
feat(horizon): opt-in horizon:listen auto-reload for development

prabowosd:
## Why

Running Laravel Horizon via `php artisan horizon` boots the app once and **caches your code**, so every edit to a job, listener, or any class a worker touches needs a manual Horizon restart (stop/start/restart in the dashboard). That stop/restart dance is exactly the friction lerd usually removes — the FrankenPHP worker entry already notes *"Code edits need `lerd restart` until we add --watch."*

Horizon ships `horizon:listen` (*"Run Horizon and automatically restart workers on file changes"*), but lerd never exposed it.

## What

A **global, opt-in** toggle `workers.horizon_reload`, surfaced three ways:

- **CLI**: `lerd horizon:reload [on|off]` (subcommand form `lerd horizon reload` too; no arg prints state).
- **Dashboard**: a System-tab "Horizon reload" panel with an on/off switch + the requirements note.
- **Config**: `workers.horizon_reload` in the global config.

When on, the Horizon worker command is **derived** into `php artisan horizon:listen --poll`. The swap happens at the single worker-command chokepoint (`WorkerStartForSite`), so the CLI, the dashboard, and the worker supervisor all pick it up on macOS and Linux. Toggling restarts running Horizon workers so it applies immediately.

## Design notes (the non-obvious parts)

- **chokidar gate, not a hard dependency.** `horizon:listen` shells out to `node bin/file-watcher.cjs` and resolves `chokidar` from the project's `node_modules`. lerd's runtime image always ships node, so the only project-side requirement is chokidar (most Laravel apps pull it in via Vite). When it's missing we **keep the standard `horizon` command and print a one-line hint** (`npm install`) rather than letting Horizon fail to boot.
- **`--poll` is mandatory.** The project is bind-mounted over **virtiofs** on macOS, where native filesystem events never reach the watcher — without `--poll` the watcher runs but never fires. Polling works on every platform.
- **Derived, not hardcoded.** The listen command is derived from the base string, so extra flags (e.g. `--environment=local`) are preserved.
- **Opt-in, default off.** Zero behaviour change for existing installs; production-like by default, lower overhead.

## Surfaces

- `internal/config/global.go` — `workers.horizon_reload` + `HorizonReloadEnabled()`/`SetHorizonReload()`.
- `internal/cli/worker.go` — `resolveHorizonCommand`/`projectHasChokidar` at the worker-command chokepoint.
- `internal/cli/horizon.go` — `lerd horizon:reload` + `ApplyHorizonReload` (persists + restarts running Horizon workers across sites).
- `internal/ui/server.go` — `horizon_reload` in `GET /api/settings` + `POST /api/settings/horizon-reload`.
- `internal/ui/web/...` — `HorizonReloadDetail.svelte` panel, `stores/horizonReload.ts`, System-tab row, i18n across all 8 locales (Indonesian translated).

## Tests

- `internal/cli/horizon_reload_test.go` — `resolveHorizonCommand` across reload off / on+chokidar / on-without-chokidar / non-horizon / flag-preservation.
- `internal/ui/web/src/stores/horizonReload.test.ts` — store load + set + failure.
- `gofmt`/`go vet` clean; `go test ./internal/cli/ ./internal/config/` green; `npm run check` + `npm test` (414 incl. locales drift) green.

## Docs

`docs/usage/queue-workers.md` gains an "Auto-reload on file changes" section.


prabowosd:
@geodro kindly can we discussion about this feature ?

geodro:
Thanks for this, the problem is real and worth solving, plenty of people hit the Horizon restart dance and horizon:listen is the right tool for it. I want to push back on the shape though before it can go in.

The main thing is that this puts framework specific behaviour in the wrong place. lerd defines its worker commands in the lerd-frameworks store yaml, the Laravel definitions for 10 through 13 already carry command: php artisan horizon, and that store file is what a real install actually runs, the Go map is only a fallback when no store version is present. This PR leaves the command in the store but then string rewrites it in core Go at the worker chokepoint, which couples the two repos together. If the store ever changes the horizon command even slightly the match silently no ops and still appends the poll flag, and you get an invalid command. Anything that knows the literal command text should live in the store, not in Go. The cleaner version is to let the Laravel yaml carry the variant, say a reload command field on the horizon worker holding php artisan horizon:listen --poll, and have core just pick base or variant based on a flag. Then all the command strings stay in the framework definition where they belong and nothing breaks when the store evolves.

Second, I don't want a dedicated Horizon reload row in the System tab. The System tab is for machine wide lerd settings, and a single framework's single worker getting its own global toggle and panel there is too specific, it doesn't generalise, and it adds surface we then maintain and translate forever. If it's exposed at all it should be a per project, worker level affordance, not a global System setting.

On the poll flag, the justification in the note is macOS specific. virtiofs only exists on macOS where podman runs inside a VM, on native Linux the bind mount is the same filesystem and inotify fires fine, so poll is just wasted CPU there and the note tells a Linux user something false about their own machine. If you keep poll it should be gated to macOS or to container mode, and the note should explain the container reason rather than naming virtiofs.

Smaller things, the locale files for de, es, fr, nl, pt and tr got the English strings verbatim so a German or French UI would show English text, only en and id are actually translated. A few of the strings and the docs section use em dashes which we don't use anywhere in the project, commas instead please. And id.json lost its trailing newline.

To be clear the feature idea is welcome, it's mostly a placement question. Move the command into the framework definition and drop the System tab setting and I think you're most of the way there.

prabowosd:
> Thanks for this, the problem is real and worth solving, plenty of people hit the Horizon restart dance and horizon:listen is the right tool for it. I want to push back on the shape though before it can go in.
> 
> The main thing is that this puts framework specific behaviour in the wrong place. lerd defines its worker commands in the lerd-frameworks store yaml, the Laravel definitions for 10 through 13 already carry command: php artisan horizon, and that store file is what a real install actually runs, the Go map is only a fallback when no store version is present. This PR leaves the command in the store but then string rewrites it in core Go at the worker chokepoint, which couples the two repos together. If the store ever changes the horizon command even slightly the match silently no ops and still appends the poll flag, and you get an invalid command. Anything that knows the literal command text should live in the store, not in Go. The cleaner version is to let the Laravel yaml carry the variant, say a reload command field on the horizon worker holding php artisan horizon:listen --poll, and have core just pick base or variant based on a flag. Then all the command strings stay in the framework definition where they belong and nothing breaks when the store evolves.
> 
> Second, I don't want a dedicated Horizon reload row in the System tab. The System tab is for machine wide lerd settings, and a single framework's single worker getting its own global toggle and panel there is too specific, it doesn't generalise, and it adds surface we then maintain and translate forever. If it's exposed at all it should be a per project, worker level affordance, not a global System setting.
> 
> On the poll flag, the justification in the note is macOS specific. virtiofs only exists on macOS where podman runs inside a VM, on native Linux the bind mount is the same filesystem and inotify fires fine, so poll is just wasted CPU there and the note tells a Linux user something false about their own machine. If you keep poll it should be gated to macOS or to container mode, and the note should explain the container reason rather than naming virtiofs.
> 
> Smaller things, the locale files for de, es, fr, nl, pt and tr got the English strings verbatim so a German or French UI would show English text, only en and id are actually translated. A few of the strings and the docs section use em dashes which we don't use anywhere in the project, commas instead please. And id.json lost its trailing newline.
> 
> To be clear the feature idea is welcome, it's mostly a placement question. Move the command into the framework definition and drop the System tab setting and I think you're most of the way there.

well noted ill implementation as you want soon thanks.

prabowosd:
Thanks, this all makes sense. Reworked along the lines you described:

**Command lives in the store, not in Go.** Added a `reload_command` field to `FrameworkWorker`; the Laravel horizon worker carries `php artisan horizon:listen`. Core no longer string-rewrites anything, it just picks `command` vs `reload_command` based on a flag. The builtin fallback carries it too, and the store side is up in geodro/lerd-frameworks#11 (Laravel 10 to 13).

**No System tab row.** Dropped the dedicated row, the detail panel, the global `workers.horizon_reload` flag and the `/api/settings/horizon-reload` endpoint. The preference is now per project (`reload_workers` in `.lerd.yaml`), surfaced as a small reload toggle on the Horizon worker itself in the site controls. CLI `lerd horizon:reload on/off` now acts on the current project.

**`--poll` gating.** You suggested the yaml hold `horizon:listen --poll`, but since you also wanted poll gated to macOS, I split the two: the store carries `horizon:listen` (framework knowledge), and core appends `--poll` only on macOS, where workers run in a container inside the podman VM and host filesystem events never reach the watcher. On native Linux the bind mount shares the host filesystem so inotify fires and poll stays off. The note explains the container reason instead of naming virtiofs. If you'd rather have the literal `--poll` in the store and strip it on Linux instead, happy to flip it.

**Smaller things.** Translated the de/es/fr/nl/pt/tr strings (were English verbatim), restored id.json's trailing newline, and removed the em dashes from the strings and docs.

PR #472, opened 2026-06-04
feat: Multi-TLD support: per-site custom endings (.test + .local + custom)

AdnanHussainTurki:
I like Lerd. I love how it tackles the restrictive nature of Herd and makes use of containers/pods to provide a clean working environment while simulating something closer to a production environment.

One limitation I encountered during migration from Herd was the lack of support for `.local` domains. I used `.local` extensively in Herd, which made the transition to Lerd somewhat frictional.

I have implemented multi-TLD support and tested it on my Mac M1. I'd appreciate your review and feedback.

## Overview

Today, every Lerd site lives under a single global `.test` ending, and that ending is effectively hardcoded in the resolver layer (the dnsmasq writer and all three Linux resolver paths). As a result, even changing the global TLD was only partially supported.

This PR makes the TLD a first-class, per-site, multi-value property. A site can now be served under multiple endings simultaneously (for example, `alice.test` and `alice.local`), while another site can continue using a single ending. Endings can be selected at link time or managed from the dashboard.

### Authoring

* Added a `--tld` flag to `lerd link` and `lerd domain add/remove`.

  * Supports comma-separated values, e.g.:

    ```bash
    lerd link alice --tld test,local
    ```
  * The first TLD is treated as the primary one and is used for the certificate CN and `APP_URL`.
* Added a new `lerd dns:apply` command to (re)apply the resolver layer for every ending across all sites. This is particularly useful after adding a new ending from the dashboard.

### Resolver Layer (Core Fix)

* `WriteDnsmasqConfigDual` now emits one `address=/.<tld>/` record per active ending instead of hardcoding `.test`.
* macOS now writes one `/etc/resolver/<tld>` file per ending and prunes stale Lerd-owned resolver files.
* All three Linux resolver paths (systemd-resolved drop-in, NetworkManager dnsmasq, and the NetworkManager dispatcher) are now templated over the active ending set.
* Diagnostics (`dns.Diagnose` / `lerd doctor`) now verify every active ending, preventing partially configured multi-TLD setups from incorrectly passing validation.

### Foundation

* Added `config.ValidateTLD` (advisory classifier):

  * Rejects multi-label endings and `localhost` in DNS-managed mode.
  * Warns about `.local` (mDNS) and real public suffixes.
* Added `config.ActiveTLDs`, which represents the union of:

  * TLDs configured across all sites, and
  * the global default TLD.

  This active set now drives the entire resolver layer.
* Fixed `.lerd.yaml` round-tripping so that non-default endings are stored correctly. A relink now restores `alice.local` instead of incorrectly producing `alice.local.test`.

## macOS Auto-Apply from the Dashboard

This PR introduces a root-owned helper (`/etc/lerd/resolver-apply.sh`) that is installed by `lerd install` and `lerd dns:apply`, along with a wildcard-free, no-argument sudoers grant.

This allows the unprivileged dashboard to enable a newly added ending without requiring a terminal step. On systems where the helper is not installed, the dashboard falls back to displaying a "run `lerd dns:apply`" message.

## Security Model (macOS Helper)

The helper is intentionally narrow in scope so that passwordless access remains low risk:

* Reads desired endings from stdin and never executes a user-writable file.
* Requires each ending to be a single DNS label, ensuring it can only modify `/etc/resolver/<label>`.
* Always writes a fixed `127.0.0.1:5300` mapping; callers cannot inject custom content.
* Uses a deny-list to reject real public suffixes as well as `local` and `localhost`, preventing rerouting of real Internet DNS traffic.
* Pruning only removes files whose content exactly matches Lerd-generated resolver files.

### Residual Trade-Off

With the helper installed, any process running as the user can configure a private development TLD without a password. This is the trade-off required to eliminate the terminal step. The deny-list prevents the mechanism from being used on real domains.

Systems that never run `dns:apply` do not install the helper and retain the existing fully manual workflow.

## Platform Notes

* Linux does not require a helper. All endings are routed through a single drop-in whose path is already covered by the existing sudoers grant, so `AutoApplyResolver` can directly invoke `ConfigureResolver`.
* `.local` is supported but with an important caveat:

  * It is owned by mDNS/Bonjour on macOS and Avahi on Linux, not by Lerd's resolver.
  * Resolution therefore depends on the operating system's handling of `.local`.
  * Health checks do not depend on `.local` resolution.
  * For reliable behavior, `.test` or another private custom label is still recommended.

## Testing

New unit tests cover:

* TLD validation
* Active TLD set generation
* dnsmasq rendering
* All three Linux resolver renderers
* Multi-TLD diagnostics
* `--tld` parsing
* `.lerd.yaml` round-tripping
* macOS helper safeguards:

  * deny-list enforcement
  * single-label validation
  * fixed-content generation
  * exact-match pruning
* Sudoers grant generation
* Web domains store
Verified locally on the macOS build configuration (-tags nogui, CGO_ENABLED=0): gofmt -l clean, and go build, go vet ./..., go test ./...  (37 packages) all pass, plus svelte-check with 0 errors. npm test is green except a pre-existing flaky notify.test.ts (not touched by this PR; passes 20/20 in isolation). New unit tests cover the validator, active-set, dnsmasq + Linux resolver renderers, multi-TLD diagnostics, --tld parsing, the  .lerd.yaml round-trip, the macOS helper safeguards, and the web domains store. The Linux CGO build/test job and the installer bats suite run on CI (not executed locally).

The full test suite passes on macOS (nogui and webui)

image


prabowosd:
good feature, but i think for development we don't need tld keep it .test this enough

AdnanHussainTurki:
@prabowosd Thanks for your comment. But I constantly feel while development `.test` is not always enough and often hits as a blocker to core functionality and testing.

Just an example I can remember is Google OAuth clients do not accept `.test` tld and for locally test the OAuth 2.0 callback, we need other TLDs. 

image


In Herd, there is option to change the TLD, but it will change the tld of all the existing and future sites, but here we have implemented it on site by site basis which I feel is quite flexible and desired by masses.

prabowosd:
> @prabowosd Thanks for your comment. But I constantly feel while development `.test` is not always enough and often hits as a blocker to core functionality and testing.
> 
> Just an example I can remember is Google OAuth clients do not accept `.test` tld and for locally test the OAuth 2.0 callback, we need other TLDs. 
> 
> image
> 
> 
> In Herd, there is option to change the TLD, but it will change the tld of all the existing and future sites, but here we have implemented it on site by site basis which I feel is quite flexible and desired by masses.
> 

Let me check deeply.

prabowosd:
Read through this end to end, including the macOS helper, and pulled it locally to trace the privileged path. Really solid work, the de-hardcoding of `.test` into `ActiveTLDs()` driving the whole resolver layer is the right shape, and the helper's security model is clearly thought through.

On the helper specifically, the escalation surface looks closed: `sudoWriteFile` does `sudo mkdir -p /etc/lerd` then `sudo tee` + `chmod 0755`, so both the directory and `resolver-apply.sh` are root-owned and not user-writable, which is what makes the no-args `NOPASSWD` grant safe. Stdin-only input, fixed file content, single-label validation, and content-matched pruning all hold up. Worth noting for reviewers that this is not a regression: the existing per-TLD `tee /etc/resolver/<tld>` grant already lets the user write arbitrary content to active-TLD resolver files, so the new helper is actually the tighter primitive.

Two things I'd raise:

1. **The deny-list lives in two places and is a deny-list.** `config.publicTLDs` (Go) and the `deny` string in `resolver_helper_darwin.go` (shell) have to be kept in sync by hand, and the comment already flags that. Since both are subsets of a ~1500-entry registry, a real-but-unlisted suffix (`.blog`, `.network`, `.name`, ...) can still be pointed at `127.0.0.1:5300`, shadowing real DNS for that suffix on the machine. The impact is local and user-initiated, so it's low severity, but the list will always lag the registry. Could the shell `deny` set at least be generated from `config.publicTLDs` so there's a single source of truth? That removes the drift even if the deny-list-vs-allow-list tradeoff stays.

2. **Maintenance/attack surface vs. scope.** This adds a permanent root-owned helper, a new sudoers grant, a deny-list to maintain, and three templated Linux resolver paths. That's a meaningful ongoing surface for a per-site multi-TLD/.local convenience. Not a blocker from me, just flagging the cost/benefit as the main thing worth a maintainer's call, especially because anything touching a passwordless privileged path raises the bar for what has to be maintained correctly forever.

Net: the implementation is careful and the security reasoning is sound. My only concrete ask is folding the two deny-lists into one source; the rest is a scope judgment for @geodro.

prabowosd:
On the Google OAuth example specifically, I went and checked the validation rules because it's a good motivating case to get right. Google requires a redirect URI to end in a public top-level domain (the console error is literally "Invalid Redirect: must end with a public top-level domain (such as .com or .org)"). That rule rejects `.test`, but it rejects `.localhost` and `.local` for the same reason, so switching a site from `.test` to `.local` doesn't actually unblock a Google OAuth client. `.local` is non-uniquely-resolvable and Google treats it the same way it treats `.test`.

The two things that do work locally for Google are:

1. The loopback exception, `http://localhost:PORT` or `http://127.0.0.1:PORT`, which Google accepts without any public-TLD check. This already works today and needs no resolver entry.
2. A real public domain you own (for example `app.mycompany.dev`) pointed at `127.0.0.1`. Google only validates that the TLD is a real public suffix, not that the host is internet-reachable.

The slightly awkward part for this PR is that the endings it makes easy (`.local`, private single-label TLDs) are the ones Google rejects, while the endings Google accepts (real public suffixes like `.com` / `.dev`) are exactly the ones `ValidateTLD` warns about and the macOS helper's deny-list refuses outright. So for the Google case as stated, multi-TLD with `.local` doesn't seem to be the lever, the loopback path covers it.

None of this is to dismiss the feature, the per-site flexibility and the Herd `.local` migration story are real and I think there's a genuine audience. I'd just suggest not leaning on the Google OAuth example as the headline justification, since loopback already solves that one. Are there providers in your workflow that reject `.test` but accept a private ending like `.local`? That's the case multi-TLD uniquely unlocks, and it'd be a stronger framing than Google.

Sources: https://developers.google.com/identity/protocols/oauth2/web-server and https://developers.google.com/identity/protocols/oauth2/resources/loopback-migration

prabowosd:
Following up on my own OAuth point, because lerd already has the better answer for it built in: `lerd share`. It exposes the current site through a public tunnel (auto-detecting ngrok, then Cloudflare Tunnel, then Expose, then localhost.run, with `--serveo` as another no-signup SSH option), and the reverse proxy rewrites `X-Forwarded-Host` and the body/headers so the framework sees the public hostname.

That gives you a real `https://...ngrok-free.app` / `https://...trycloudflare.com` URL, which is exactly what Google's redirect-URI validation wants (public TLD + https). So the Google OAuth callback case is already solved today, regardless of a site's local ending, and it works for the providers that reject `.test` precisely because the tunnel hostname is a genuine public domain.

Docs: https://geodro.github.io/lerd/usage/sites#sharing-sites

So I'd gently retract the OAuth framing entirely as a justification for multi-TLD: loopback covers the simple case and `lerd share` covers the public-callback case. The per-site flexibility and the Herd `.local` migration story still stand on their own as the motivation, those just feel like the stronger ground than OAuth.

prabowosd:
And to be clear, none of the above is me arguing against the PR. The work is genuinely good and worth landing: the per-site multi-TLD model is more flexible than Herd's global switch, the resolver de-hardcoding is clean, and the security model behind the helper is carefully done. My comments are just sharpening the framing and the one deny-list-dedup nit, not blocking it.

Thanks for putting this together and for the thoughtful back-and-forth, @AdnanHussainTurki. Nice contribution.

geodro:
Thanks for this, both of you. Adnan, this was a careful and well thought out piece of work, the per site model is genuinely more flexible than Herd's global switch and the resolver refactor is clean, so please don't read the close as the idea not landing. And prabowosd, thanks for the depth you put into the review, that kind of tracing and honest framing is exactly what makes these discussions worth having.

The per site endings idea and the Herd .local migration story are real and there's an audience for them, so I'd be glad to see this come back in a more scoped form down the line. Thanks again to both of you for the thoughtful back and forth on it.

PR #473, opened 2026-06-04
fix(macos): pin --platform on pull for amd64-only images, register only after pull succeeds

geodro:
## Problem

On Apple Silicon, installing Postgres 18 from the dashboard fails and leaves a broken service behind:

```
pulling docker.io/postgis/postgis:18-3.6-alpine: exit status 125
```

```
Error: ... choosing an image from manifest list docker://postgis/postgis:18-3.6-alpine:
no image found in image index for architecture "arm64", variant "v8", OS "linux"
```

Two things go wrong: the pull fails, and the service gets registered anyway, so it shows as installed but never starts.

## Root cause

`postgis/postgis` publishes no arm64 manifest for any tag (and neither does `mysql:5.7`; `8.4` and `9.7` do). The quadlet already pins `--platform=linux/amd64` so the container runs under the podman machine's emulation, but the explicit pull step the streaming install added to surface progress ran a bare `podman pull` with no platform. On arm64, podman cannot pick an arch from the manifest list and exits 125, so pull and run disagreed.

Separately, `InstallPresetStreaming` (`internal/serviceops/serviceops.go`) wrote the YAML config and quadlet **before** pulling the image. Quadlet presence is the install-state truth (`ServiceInstalled`), so a failed pull left the service registered with no working image.

## Fix

- **Pin the platform on pull.** New `PlatformPullArgs(image)` injects `--platform=linux/amd64` for the no-arm64 images on darwin, no-op on linux. Every pull path routes through it: the three pull functions in `internal/podman/exec.go`, the FPM base pull in `build.go`, and the two CLI start/warm-up pulls in `startstop.go`. Both `PlatformPullArgs` and the run-time `PlatformPodmanArgs` now key off a single `imageLacksArm64(image)` predicate, so the quadlet's `--platform` and the pull always agree regardless of the service's name.
- **Pull before register.** `InstallPresetByName` is split into `resolvePresetForInstall` (validate, no writes) and `registerPreset` (save config + quadlet). The streaming flow now resolves, pulls, registers, then starts dependencies and the unit. A failed pull writes nothing; registration still lands before the dependency-start loop, so a dependency that fails to start leaves the service installed on disk (this matters on reinstall, which removes the prior copy first). `InstallPresetByName` stays a thin `resolve + register` wrapper, so the CLI, MCP, and link callers are unchanged.
- `mysql:5.7` had the same amd64-only gap as postgis and is covered by the same predicate.

## Linux is unaffected

`PlatformPullArgs` lives behind the build tag: `platform_linux.go` returns nil, so `pullArgs` produces the same `["pull", image]` it always did. The register-before-pull reorder is OS-agnostic and is the intended correctness fix on every platform: a failed pull should never register a service.

Self-reported (no separate issue); happy to open one if you'd prefer to track it there.

### DIFF
--- a/internal/cli/startstop.go
+++ b/internal/cli/startstop.go
@@ -114,7 +114,8 @@ func ensureImages() {
 			jobs = append(jobs, BuildJob{
 				Label: "Pulling " + label,
 				Run: func(w io.Writer) error {
-					cmd := podman.Cmd("pull", label)
+					args := append(append([]string{"pull"}, podman.PlatformPullArgs(label)...), label)
+					cmd := podman.Cmd(args...)
 					cmd.Stdout = w
 					cmd.Stderr = w
 					return cmd.Run()
@@ -671,7 +672,8 @@ func startRestoredServices() {
 		pullJobs = append(pullJobs, BuildJob{
 			Label: "Pulling " + img,
 			Run: func(w io.Writer) error {
-				cmd := podman.Cmd("pull", img)
+				args := append(append([]string{"pull"}, podman.PlatformPullArgs(img)...), img)
+				cmd := podman.Cmd(args...)
 				cmd.Stdout = w
 				cmd.Stderr = w
 				return cmd.Run()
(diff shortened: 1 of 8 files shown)


prabowosd:
Oh yeah i figure this issue too on my macbook great @geodro 🤙🏻

PR #476, opened 2026-06-05
fix(start): auto-heal podman overlay corruption + lerd machine reset

### LINKED ISSUE #475: Lerd fails to start if not stopped prior to shutdown
OS: macOS 26
Lerd: 1.23.1
Podman: 5.8.2, 1.27.2 (Desktop)

I've had this issue happen a couple of times so I don't believe its coincidence anymore.

Basically when I shutdown my computer while Podman and Lerd are still running and then start it again, `lerd start` now fails when starting with the following:

```
✗ nginx: podman run lerd-nginx: exit status 125: Error: getting graph driver info "b77616fb55faa74c53cc0582eb915e57c0afc85c636e3e0d94be23f2e7f4e945": readlink /var/lib/containers/storage/overlay: invalid argument
✗ php85-fpm: podman run lerd-php85-fpm: exit status 125: Error: getting graph driver info "7614dcc4b120e7b2eb812027c4f3c3a79648842f9714eb1ec2062be392d78924": readlink /var/lib/containers/storage/overlay: invalid argument
✗ mailpit: podman run lerd-mailpit: exit status 125: Error: getting graph driver info "eb37f5662d62409ae77a235a9679f0483c79b53877b9e3581bdb18a7cf019b06": readlink /var/lib/containers/storage/overlay: invalid argument
✗ mysql: podman run lerd-mysql: exit status 125: Error: getting graph driver info "d4d22ac5763bc7836a0d2109f05c032b98a03d7f6e21f8502beb5821c17a60d5": readlink /var/lib/containers/storage/overlay: invalid argument
✗ redis: podman run lerd-redis: exit status 125: Error: getting graph driver info "60d67dd230f52ad6934e08eda9c8e258a6ea7045018a25d90ac8c60a3a8a6124": readlink /var/lib/containers/storage/overlay: invalid argument
```

Running `podman run` manually seems to work however so I'm unclear on what the issue is here.

The only solution to this problem has been to delete the Podman Machine and recreate it.

If I run `lerd stop` and close Podman Desktop prior to my shutdown, then when I start my computer again I don't experience this issue.

prabowosd:
## Problem

Closes #475.

After an ungraceful macOS host shutdown while the Podman Machine VM is still running, the VM's container storage is left with a stale overlay mount and **every** container start fails:

```
exit status 125: Error: getting graph driver info "<id>": readlink /var/lib/containers/storage/overlay: invalid argument
```

`podman run` manually works; the only known fix was deleting & recreating the machine. `lerd start` surfaced the raw podman error with no recovery path.

## Why recovery is safe in lerd

lerd bind-mounts all persistent data to the **host** (`internal/podman/quadlet_generate.go` → `Volume=<config.DataSubDir(name)>:/var/lib/mysql`), not into in-VM named volumes. So restarting **or** recreating the VM is non-destructive: databases and site data survive, and images are rebuilt by the existing "ensure images built" preflight. This makes auto-heal feasible where it wouldn't be in a generic podman setup.

## Changes (macOS; Linux uses native rootless podman, no VM)

1. **Auto-heal in `runStart`** — detect the overlay-storage error from the service start pass, restart the Podman Machine to remount its storage (reusing `ensurePodmanMachineRunning`), and retry the start pass **once**. If it still fails, print recovery guidance instead of the cryptic podman error.
2. **`lerd machine reset`** — recreate the VM (stop → `machine rm -f` → re-init) with a confirmation prompt (skip via `--yes`), automating the manual workaround safely.

The error matcher (`isOverlayStorageError`) lives in a platform-agnostic file so it's unit-tested on the Linux CI runner. The heal and reset command are darwin-only with non-darwin no-op stubs.

## Design notes

- **Single retry, no loop** — the heal restarts the VM at most once per `lerd start` to avoid a restart loop on genuinely unrecoverable storage.
- **Tier 2 is manual, not auto-recreate** — auto-recreating the user's machine would also drop non-lerd containers/images (e.g. Podman Desktop usage), so recreation stays behind an explicit, confirmed `lerd machine reset`.
- The heal refreshes the machine-restart baseline (`recordMachineLastUp`) so the next run doesn't mistake the heal's restart for an external one.

## Tests

`TestIsOverlayStorageError` — the real #475 string matches; port conflicts, missing-image, and unrelated "invalid argument" errors do not. Runs on the Linux CI runner (matcher is platform-agnostic).

## Verification

- `go build ./...`, `go vet ./internal/cli/... ./cmd/...`, and `go test ./internal/cli/ -run TestIsOverlayStorageError` all green on darwin.
- `lerd machine` / `lerd machine reset --help` render the expected command tree.

### DIFF
--- a/cmd/lerd/main.go
+++ b/cmd/lerd/main.go
@@ -115,6 +115,7 @@ func main() {
 	root.AddCommand(cli.NewWhatsnewCmd())
 	root.AddCommand(cli.NewManCmd())
 	root.AddCommand(cli.NewDoctorCmd())
+	root.AddCommand(cli.NewMachineCmd())
 	root.AddCommand(cli.NewBugReportCmd())
 	root.AddCommand(cli.NewLogsCmd())
 	root.AddCommand(cli.NewOpenCmd())
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -376,7 +376,7 @@ readlink /var/lib/containers/storage/overlay: invalid argument
 
 Cause: the macOS host was shut down ungracefully (forced power-off, battery death, kernel panic) while the Podman Machine VM was still running. The VM's container storage is left with a stale overlay mount and corrupt container layers, so no container can start until the storage is remounted and the stale containers are rebuilt.
 
-`lerd start` detects this and **self-heals automatically** on the first run: it restarts the Podman Machine to remount the storage, force-removes the stale `lerd-*` containers so they rebuild on fresh storage, and retries the start pass once. Your data is safe throughout — lerd bind-mounts every database and site directory to the host, not into the VM.
+`lerd start` detects this and **self-heals automatically** on the first run: it restarts the Podman Machine to remount the storage, force-removes the stale `lerd-*` containers so they rebuild on fresh storage, and retries the start pass once. Your data is safe throughout: lerd bind-mounts every database and site directory to the host, not into the VM.
 
 If the automatic recovery isn't enough (it prints guidance pointing here), recreate the VM:
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -365,3 +365,24 @@ cat "${XDG_RUNTIME_DIR:-/run/user/$(id -u)}/containers/networks/aardvark-dns/ler
 # if only 10.89.7.1 is present, the drift fix didn't run — re-run lerd install
 ```
 :::
+
+::: details Podman Machine overlay-storage error (macOS)
+Symptom: on macOS, `lerd start` fails and **every** container start reports a graph-driver / overlay error:
+
+```
+exit status 125: Error: getting graph driver info "<id>":
+readlink /var/lib/containers/storage/overlay: invalid argument
+```
+
+Cause: the macOS host was shut down ungracefully (forced power-off, battery death, kernel panic) while the Podman Machine VM was still running. The VM's container storage is left with a stale overlay mount and corrupt container layers, so no container can start until the storage is remounted and the stale containers are rebuilt.
+
+`lerd start` detects this and **self-heals automatically** on the first run: it restarts the Podman Machine to remount the storage, force-removes the stale `lerd-*` containers so they rebuild on fresh storage, and retries the start pass once. Your data is safe throughout — lerd bind-mounts every database and site directory to the host, not into the VM.
+
+If the automatic recovery isn't enough (it prints guidance pointing here), recreate the VM:
+
+```bash
+lerd machine reset
+```
+
+This stops the VM, removes it, and re-initialises it. Databases and site data are preserved (they live on the host); container images are rebuilt automatically on the next `lerd start`. See [Start, Stop & Autostart → `lerd machine reset`](usage/lifecycle.md#lerd-machine-reset-macos).
+:::
--- a/docs/usage/lifecycle.md
+++ b/docs/usage/lifecycle.md
@@ -89,6 +89,23 @@ The system tray's **Quit Lerd** menu item calls `lerd quit`.
 
 ---
 
+## `lerd machine reset` (macOS)
+
+```bash
+lerd machine reset        # asks for confirmation first
+lerd machine reset --yes  # skip the prompt
+```
+
+Recreates the Podman Machine VM. Reach for it only when `lerd start` reports a container-storage error such as `getting graph driver info ... overlay: invalid argument`, which happens after the macOS host is shut down ungracefully while the VM is still running and leaves the VM's container storage corrupt. See [Troubleshooting → Podman Machine overlay-storage error](../troubleshooting.md).
+
+The command stops the VM, removes it (`podman machine rm -f`), and re-initialises it. **Your data is preserved:** lerd bind-mounts every database and site directory to the host, not into the VM, so only the VM's container storage and images are discarded. Images are rebuilt automatically on the next `lerd start`.
+
+::: tip lerd start already tries to self-heal
+On macOS, `lerd start` detects this exact error and attempts an automatic recovery first (remount the VM's storage, rebuild the stale containers, retry once). `lerd machine reset` is the manual fallback for when that recovery isn't enough. This command is macOS-only; Linux runs podman natively with no VM.
+:::
+
+---
+
 ## Autostart on login
 
 Lerd can boot itself every time you log in. Autostart is a single switch over every lerd-owned systemd user unit on the machine:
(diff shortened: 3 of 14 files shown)


geodro:
I went through the whole change, traced how the podman run failure on macOS actually bubbles up through RunParallel into the matcher, and built and tested it locally. The structure is solid and the matcher is nicely conservative, it wants the graph driver phrase plus an overlay or readlink signal, and the negative cases for port conflicts and missing images mean it won't fire on unrelated failures. Reusing the existing machine plumbing and refreshing the LastUp baseline after the restart so the gvproxy heal doesn't double fire next run is a nice touch.

My main hesitation is the same thing Sn0wCrack flagged at the end of the issue, he tried restarting the machine once and it didn't clear the error, only a full reset did. The primary heal path here is exactly a machine stop, start, and one retry, so in the cases that actually bite, the user waits out a slow machine restart, the retry fails again, and then we tell them to run lerd machine reset anyway. It's a safe fallback so I'm not worried about it doing harm, but it may not deliver the heal it advertises. The clue is that a fresh podman run works for him because new storage gets allocated, which points at the existing container or image layer being the corrupt part rather than the overlay base that a remount fixes. Worth noting the heal path doesn't force remove the stale lerd containers the way the gvproxy heal does, and replace itself may hit the same getting graph driver info error while removing them, so neither the restart nor the retry necessarily gets us to the fresh storage that's known to work. I'd hold merge until you get a confirmation from the branch test, otherwise we're shipping a thirty second detour that lands on the manual workaround regardless.

A few smaller things. When someone declines the confirmation prompt in machine reset we return an aborted error, so a normal no makes cobra print Error: aborted and exit non zero, that should just print cancelled and return nil since choosing not to proceed isn't a failure. The recovery guidance prints before the worker start pass, so when services are still down every worker also fails right afterwards and buries the one line that tells the user how to recover, either skip the worker pass when services failed this way or print the guidance last. The retry reruns the entire service pass and with replace that restarts already healthy containers, which is moot in the overlay case but is churn if the matched error ever came from a single service. The nolint errcheck on the retry is misleading too since serviceErr is assigned and then consumed by the outcome reporter, so it isn't actually unchecked.

Last thing, this adds a new lerd machine reset command and a new recovery flow but doesn't touch docs, can you add a machine reset entry to the lifecycle page and a troubleshooting section for the overlay error pointing at the reset, leaving the changelog alone since this isn't a release.

prabowosd:
Thanks for the deep trace @geodro — pushed 5006d26 addressing everything.

**Main concern (the heal may not deliver).** You nailed it: Sn0wCrack's "restart didn't clear it, only a reset did" plus "a fresh `podman run` works" points at the corruption living in the existing container/image layers, not just the overlay base a remount fixes. So the heal now does what the gvproxy heal does — after the machine restart it **force-removes the stale `lerd-*` containers** so the retry's `podman run` allocates fresh container storage, which is the path that's known to work. I refactored the gvproxy removal into a shared `forceRemoveLerdContainers(includeStopped, announce)`; the overlay heal passes `includeStopped=true` (`-a`) because the failed containers never reached running state, unlike the gvproxy case. If `rm -f` itself hits the graph-driver error we log a WARN and the retry still runs; if that fails too we land on the `lerd machine reset` guidance as the safe fallback.

That said — agreed this needs a real branch-test confirmation before merge. I can reproduce the stale-overlay state locally and verify the rebuild actually recovers; **holding merge until I post that result here.** If the container-rebuild still doesn't clear it on a real corrupt VM, then reset is genuinely the only path and I'll demote the auto-heal to just detect-and-guide.

**Smaller items, all in:**
- `machine reset` declining the prompt now prints `Cancelled.` and returns nil (no more `Error: aborted` + non-zero exit).
- Recovery guidance is no longer buried: when the overlay error persists after the retry, `runStart` skips the worker pass *and* the DNS/tray steps and prints the guidance last, then returns.
- Dropped the misleading `//nolint:errcheck` on the retry — `serviceErr` is assigned and consumed by `isOverlayStorageError`/`reportOverlayHealOutcome`.
- Re: the retry rerunning the whole service pass — left as-is since it's moot in the overlay case (all services are down, so a full re-run is correct); happy to scope it per-failed-unit if you'd rather.

**Docs:** added a `lerd machine reset` section to the lifecycle page and a troubleshooting entry for the overlay-storage error pointing at it. Left the changelog alone.

prabowosd:
## Branch test — reproduced overlay corruption and verified recovery

Spun up a throwaway Podman Machine (applehv only allows one VM active, so this ran with the working machine stopped), broke a container's RW overlay layer while leaving the image layers intact, and walked the recovery levels:

```
# (A) existing container with the broken layer:
$ podman start lerd-redis
Error: unable to start container "3752898bf744": mounting storage for container 3752898bf744:
readlink /var/lib/containers/storage/overlay/312d216688349/diff: no such file or directory   # exit 125

# (B) fresh run — new RW layer, same intact image:
$ podman run -d --name lerd-fresh redis:alpine
e12761a2537a...   # exit 0  ✅

# recovery: rm -f the broken containers, then recreate (= the heal's rebuild + retry):
$ podman rm -f lerd-redis lerd-nginx        # exit 0  ✅  (rm works on the corrupt containers)
$ podman run -d --name lerd-redis redis:alpine && podman run -d --name lerd-nginx nginx:alpine
# both exit 0, all three containers running  ✅
```

**Conclusion on your main concern:** the heal does deliver. When the image layers are intact (Sn0wCrack's "a fresh `podman run` works" tell), the corruption is in the container layers, and `podman rm -f` + recreate recovers them fully — which is exactly what `forceRemoveLerdContainers` + the retry now do after the machine restart. The 30-second detour lands on a working stack, not on the manual workaround. (Caveat I can't fully close: I induced the broken-layer state surgically rather than via a real host power-loss, since the latter is non-deterministic. If a real corrupt VM ever resists the rebuild, the reset guidance is still the safe fallback.)

**One thing the repro caught — a matcher gap (pushed in 4e98f4c).** My reproduced variant reports `mounting storage for container ...: readlink .../overlay/<layer>/diff: no such file or directory` — no "graph driver" phrase, so the original matcher missed it. I broadened `isOverlayStorageError` to also recognise the container-mount variant, still gated on the overlay-store path **and** a readlink failure so port conflicts / missing images / generic mount errors stay out. Added positive + negative test cases (10 pass). If you'd rather keep the matcher strictly scoped to the exact #475 graph-driver string, say the word and I'll drop that second branch — the heal logic itself doesn't depend on it.

Net: 4e98f4c (matcher + tests) on top of 5006d26 (rebuild stale containers, prompt/guidance/nolint fixes, docs). I think this is mergeable now, but it's your call on the matcher breadth.

PR #497, opened 2026-06-09
Group the MCP tool surface and add Codex, Gemini, and Copilot

geodro:
This refactors lerd's MCP integration in three connected ways: it makes client registration table driven, consolidates the tool surface, and sources all assistant docs from one place. Along the way it adds support for Codex CLI, Gemini CLI, and GitHub Copilot.

Client registration used to be open coded per assistant, so adding one meant duplicating write and remove logic across four functions. There is now a single aiClients table describing each assistant's config path, format, and context files, with generalized writers for the three formats in play: the mcpServers JSON key (Claude, Cursor, Junie, Windsurf, Gemini), VS Code's servers key with a stdio type (Copilot), and Codex's TOML. Claude global registration stays on its own claude mcp add CLI because that file holds all of Claude's user state.

The server exposed around eighty flat tools, so every session carried that entire manifest and the set changed shape mid session depending on the detected framework. Those are now ten resource groups, site, service, db, env, runtime, worker, exec, framework, diag, and worktree, each selected by an action argument and routed through one dispatch table that reuses every existing handler unchanged. Because the manifest is static the prompt cache stays stable, and the tools/list payload dropped from about thirty two thousand bytes to about sixteen thousand. This is a clean break: the old flat tool names no longer exist, so callers use the group with the matching action instead.

The three large hand maintained doc constants collapse into one embedded reference that every client renders from with its own thin frontmatter, so the guidance can no longer drift between assistants.

Codex config.toml is edited textually rather than round tripped through a TOML encoder, which would otherwise strip a user's comments and reorder their tables. On update only the files a client was already set up with are refreshed, so a routine upgrade never drops new client files into a repo or a home directory for an assistant the user never enabled; to pick up a newly supported assistant, re-run mcp:enable-global or mcp:inject.


prabowosd:
I think will be nice if we have for antigravity cli too

geodro:
> I think will be nice if we have for antigravity cli too

Good idea!

prabowosd:
Let's add to lerd 😎

geodro:
> Let's add to lerd 😎

https://github.com/geodro/lerd/pull/498

PR #507, opened 2026-06-10
feat(sites): show Laravel APP_NAME on the dashboard tiles

prabowosd:
## What

The sites overview dashboard identifies each site only by its URL. Laravel projects already declare a human-friendly `APP_NAME` in their `.env`, so this surfaces it on the tile.

- **Backend** (`internal/ui/server.go`): new `app_name` field on `SiteResponse`, populated via a small `laravelAppName(frameworkName, sitePath)` helper that reads `APP_NAME` with `envfile.ReadKey` — **only for Laravel sites**. Non-Laravel projects, a missing `.env`, or an absent `APP_NAME` all yield `""` (the field is `omitempty`).
- **Frontend** (`SiteTile.svelte` + `sites.ts`): when `app_name` is present it becomes the tile title and the domain drops to the subline (`MyApp` / `myapp.test · Laravel · PHP 8.3`). Non-Laravel tiles keep the domain as the title and render exactly as before.

## Why

The URL is the machine identity; the app name is how people actually refer to a project. Showing it makes the overview scannable when you run several Laravel sites whose domains look alike.

## Testing

- `go test ./internal/ui/` — added `TestLaravelAppName` (laravel / non-laravel / missing `.env` / empty path / absent `APP_NAME`).
- `vitest` — added a `SiteTile` case asserting the app_name title + domain subline.
- `svelte-check`: 0 errors. `gofmt`: clean.
- Built locally (`make install`) and verified against real sites: Laravel sites report their `APP_NAME` (e.g. `LaravelSite1`, `LaravelSite2`), a CodeIgniter site reports `""`.

Additive and backwards compatible — no migration, no config, no CLI surface change.


geodro:
This is really nice, I pulled it down and tried it locally and the tiles read much better with the app name than the bare URL. One small thing before it goes in. A fresh Laravel install ships with APP_NAME set to Laravel and a lot of people never change it, so on a dashboard with a few uncustomised sites they would all show up titled Laravel, which is actually less scannable than the domains they replace. Could you also return empty when APP_NAME is the default Laravel, case insensitive, so the tile only takes over the title when the name is genuinely customised and otherwise falls back to the domain. That keeps it purely additive. Everything else looks solid, the Laravel only gating and the tests are spot on. Thanks for this.

prabowosd:
Good catch, done in 54a6baf. `laravelAppName` now returns "" when APP_NAME is still the stock `Laravel` default (case-insensitive via `strings.EqualFold`), so uncustomised sites fall back to the domain and the label stays purely additive. Added cases to `TestLaravelAppName` covering `Laravel`/`laravel`/`LARAVEL`.

PR #509, opened 2026-06-10
feat(sites): add a Laravel Doctor panel to the site detail view

prabowosd:
## What

A new **Doctor** tab on the site detail view (Laravel sites only) that runs app-level health checks and offers one-click fixes. The panel fills the gap between the Env tab and the command runner: it surfaces the misconfigurations that actually bite during local dev and wires each to the command that resolves it.

Checks (`GET /api/sites/{domain}/doctor`, loopback-only):

| Check | What it flags | Fix |
|---|---|---|
| `app_key` | `APP_KEY` empty | `key:generate` |
| `env_drift` | keys in `.env.example` missing from `.env` (names only, never values) | — |
| `app_debug` | `APP_DEBUG` on while `APP_ENV=production` | — |
| `storage_link` | missing `public/storage` symlink when the public disk is in use | `storage:link` |
| `migrations` | pending migrations via `migrate:status`; degrades to **unknown** when the app/DB is down | `migrate` |

The cheap checks read files only and run inline; only `migrations` execs in the container (same trust level + PATH shimming as the command runner, hence loopback-only). Findings carry an optional `fix` naming a command from the site's command set — the **Fix** button runs it through the existing command runner (`executeCommand` → `CommandRunModal`) and re-checks. The endpoint adds **no mutation of its own**.

To make the fixes runnable, `key:generate` and `storage:link` are now exposed as Laravel framework commands alongside the existing `migrate` (so they also show up in the command palette/dropdown).

## Why

The URL/env/tinker tabs tell you how a site is wired; nothing told you *what's wrong* with it. "Pulled main and the app 500s on a new env var", "forgot to migrate", "uploads 404 because storage:link never ran" are the daily papercuts — this turns each into a visible finding with a one-click fix.

## Implementation notes

- Backend: `internal/ui/site_doctor.go` (checks + `doctorRoute`), registered in `handleSiteAction`. The check status shape (`{name, status, detail}`) mirrors the MCP doctor for consistency. The migrations check uses the same `sh -c` + `BinDir()` PATH shimming as `handleCommandRun`.
- Frontend: `SiteDoctorTab.svelte` + `stores/doctor.ts`, wired into `SiteDetail.svelte` as a tab gated on `is_laravel`.
- Non-Laravel sites return empty checks; the tab isn't shown for them.

## Testing

- Go: `internal/ui/site_doctor_test.go` covers each cheap check (app_key / env_drift / app_debug / storage_link) and the `migrate:status` parser across temp-dir fixtures. `go test ./internal/config/` (371) and `./internal/ui/` green; `gofmt`/`go vet` clean; full `go build` OK.
- Web: `SiteDoctorTab.test.ts` (renders checks + Fix button, omits Fix when no command matches). Full suite **513/513**, `svelte-check` 0 errors, locale drift test passes (17 keys × 14 locales).
- Built locally (`make install`) and verified against real sites: `lferp.test` reports pending migrations (fix=migrate), `basepos.test` flags a missing storage link (fix=storage:link), a CodeIgniter site returns `checks: []`.

No migration, no config, no new dependencies. i18n: en + id translated, the other 12 locales seeded with English (happy to defer to your i18n pass).

Docs: a Doctor tab bullet added to `docs/features/web-ui.md`.


geodro:
Really like this, the checks are practical and the storage link and env drift catches are the kind of thing that bites silently. Two bits of feedback before it goes in.

On placement, I would move it out of a dedicated tab. The doctor is something you glance at and act on rather than a view you sit in, so a permanent tab on every Laravel site adds weight even when everything is healthy. I think it fits better as a button in the site overview, right next to the commands button, that opens a modal with the results. That keeps it on demand, matches how the other transient actions in lerd already work as modals, and lets you hang a small badge on the button later showing the failure or warning count so it only draws attention when there is something to fix. The migrate:status exec then runs on an explicit click rather than every time someone switches to the tab.

The other thing is the translations. The non English locale files have the new doctor keys but the values are still English, so de, es, fr, ja, zh, ro, it, nl, pl, pt, tr and vi all render in English, only the Indonesian file is actually translated. Could you drop the changes to the non English message files and ship just en.json, I will run them through the translation flow afterwards so those locales do not regress with untranslated strings.

Nice work overall, the backend is clean and routing the fixes through the existing command runner instead of new mutation endpoints is exactly the right call.

prabowosd:
Both addressed in f7bf805.

**Placement:** the doctor is no longer a tab. There's now a **Doctor** button in the overview controls right next to Commands that opens a modal (`SiteDoctorModal.svelte` on the shared `Modal` base, matching the other transient-action modals). The checks, including the `migrate:status` exec, run only when the modal opens, so a healthy site carries no extra weight and the exec fires on an explicit click. Removed the tab wiring from `SiteDetail` and the `sites_tabs_doctor` key. I left the badge off for now since a live count would mean running the checks eagerly, which is exactly what we're avoiding, but the count is right there in the report when you want to wire it later.

**Translations:** reverted all non-English message files to baseline; only `en.json` carries the new keys now, so the other locales fall back to English instead of regressing. All yours for the translation flow.

PR #510, opened 2026-06-10
feat(sites): Clear logs button to reclaim disk in the App Logs tab

prabowosd:
## What

The App Logs tab now shows the **total log size** and a **Clear logs** button that deletes the project's declared log files to reclaim disk (`POST /api/app-logs/{domain}/clear`, loopback-only).

Laravel projects pile up large `storage/logs/*.log` files during local dev that nothing ever cleans up. This adds a one-click reclaim right where you already look at those logs.

## How

- The sweep enumerates files through the **same traversal-guarded discovery the viewer uses** (`applog.DiscoverLogFiles` + `ResolveLogFilePath`), so it can only ever touch the framework's declared log globs. A non-log file in `storage/logs/` is left untouched.
- The active log is recreated by the app on its next write, so this is self-healing — hence a lightweight **two-step confirm** on the button rather than a modal.
- A per-file delete error is remembered but doesn't stop the sweep, so one locked file can't strand the rest.

## Testing

- Go: `clearAppLogs` unit test — deletes the matched `*.log` files, counts bytes freed, leaves a sibling `.gitignore` untouched, and a second sweep is a clean no-op.
- Web: a new `AppLogsTab` case asserts the two-step confirm — the first click only arms, the second `POST`s to `/clear`. Full suite **514**, `svelte-check` 0 errors, locale drift test passes (4 keys × 14 locales).
- Built locally (`make install`) and verified against real sites.

No migration, no config, no new dependencies. i18n: en + id translated, the other 12 locales seeded with English.

Docs: the Clear logs button is noted on the App logs tab bullet in `docs/features/web-ui.md`.


geodro:
Tested this and it works exactly as intended, I cleared a site's logs and it freed precisely the right number of bytes and only touched the files inside storage/logs, reusing ResolveLogFilePath for the traversal guard is the right call. Genuinely useful too, some of my laravel.log files had grown to tens of megabytes.

One change on the confirmation. I would make it a modal rather than the inline two step button. Clearing deletes real log files, and even though the app recreates them, a quick double click wipes a lot of history in one go, so it deserves a deliberate confirm. A modal also keeps it consistent with how lerd confirms its other destructive actions instead of introducing a new pattern.

Same translation note as the doctor PR, the non English message files have the four new keys but the values are still English for de, es, fr, ja, zh, ro, it, nl, pl, pt, tr and vi, only the Indonesian file is translated. Could you drop the non English message changes and ship just en.json, I will run them through the translation flow afterwards.

Otherwise this is solid, the backend is safe and the byte reporting is accurate.

prabowosd:
Both addressed in 8bbb8cf.

**Confirmation:** dropped the inline two-step button for a confirmation modal (`ClearAppLogsModal.svelte` on the shared `Modal` base, with a danger-toned confirm). The Clear logs button now just opens the modal showing the total size, and the delete only fires on an explicit confirm, consistent with lerd's other destructive-action modals. Updated the AppLogsTab test to assert nothing is sent until the modal is confirmed.

**Translations:** reverted all non-English message files to baseline; only `en.json` carries the new modal keys now, so the others fall back to English instead of regressing. Yours for the translation flow.

PR #512, opened 2026-06-10
test(status): add bun fields to dnsState base fixture

prabowosd:
The php-containerfile change (#511) added `bun_available`, `bun_version` and `using_system_bun` to `StatusResponse`, but the `dnsState` test fixture in `status.test.ts` was not updated to match. As a result `npm run check` (svelte-check) fails on `main` with five errors, all in `src/stores/status.test.ts`, because the `base` object is missing those three required fields.

This adds the three fields (`bun_available: false`, `bun_version: ''`, `using_system_bun: false`) to the fixture so the typecheck goes green again. Test-only change, no runtime impact.

Before:
```
COMPLETED 1408 FILES 5 ERRORS 0 WARNINGS 1 FILES_WITH_PROBLEMS
```
After:
```
COMPLETED 1408 FILES 0 ERRORS 0 WARNINGS 0 FILES_WITH_PROBLEMS
```

### DIFF
--- a/internal/ui/web/src/stores/status.test.ts
+++ b/internal/ui/web/src/stores/status.test.ts
@@ -108,7 +108,7 @@ describe('status store', () => {
 
   it('dnsState reads the status field and falls back to ok for old payloads', async () => {
     const { dnsState } = await import('./status');
-    const base = { nginx: { running: true }, php_fpms: [], php_default: '', node_default: '', node_managed_by_lerd: true, watcher_running: true };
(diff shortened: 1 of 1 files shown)


prabowosd:
Feel free closes this PR if already fixed in your local @geodro

geodro:
Thanks for catching this, but it is already resolved on main. The same three bun fields were added to the dnsState base fixture as part of PR 509, which merged earlier today, so svelte-check is green on main now. This PR is currently conflicting against that change and would only reorder the same fields, so closing it as superseded.

Issue #530, opened 2026-06-15
Rootless containers lose internet after host network change/suspend; no recovery and lerd stop leaves php-fpm running


D90-dev:
_Revised: the original report blamed host DNS (`~.` routing). That was wrong. The real problem is container egress, described below._

### Summary

lerd runs `composer` and `php` inside the project's rootless Podman PHP-FPM container. When the host network changes while lerd is running (Wi-Fi reconnect, DHCP renew, or suspend and resume), the containers lose all outbound internet and do not recover on their own. Anything run in a container that needs the internet then fails. With composer the first symptom is a DNS error, because resolving a host is the first thing it does:

```
curl error 6 while downloading https://repo.packagist.org/p2/aws/aws-sdk-php.json: Could not resolve host: repo.packagist.org
```

The host itself resolves names and reaches the internet fine the whole time. The failure is entirely inside the container network.

### Root cause

The rootless network namespace that backs the bridge network keeps its slirp4netns uplink only from the moment the first container started. It does not follow later host network changes. After such a change the namespace loses its default route and tap device, while a stale slirp4netns process keeps running, so containers can still reach each other but have no path out.

Broken state (container egress dead):

```
$ podman unshare --rootless-netns ip route
10.89.0.0/24 dev podman1 proto kernel scope link src 10.89.0.1
# no default route, no tap0

$ podman exec <php-container> wget -qO- https://1.1.1.1/
wget: can't connect to remote host (1.1.1.1): Network unreachable

# container-to-container name resolution still works, which hides the problem
$ podman exec <nginx-container> nslookup <other-container> 10.89.0.1
Address: 10.89.0.3
```

After a full restart the uplink is rebuilt and egress works again:

```
$ podman unshare --rootless-netns ip route
default via 10.0.2.2 dev tap0
10.0.2.0/24 dev tap0 proto kernel scope link src 10.0.2.100
10.89.0.0/24 dev podman1 proto kernel scope link src 10.89.0.1
```

### Two lerd-side problems

1. No recovery after a host network change. The containers stay broken until every one of them is stopped and started again. A reboot is not required, but nothing in lerd notices the change or rebuilds the uplink.
[...]

PR #532, opened 2026-06-15
fix(stop): retry systemd stop jobs reported "canceled" so lerd stop leaves no container running

prabowosd:
Addresses the second of the two `lerd`-side problems in #530: **`lerd stop` does not stop the PHP-FPM container cleanly** (`stop lerd-php85-fpm failed: canceled`, container left running). The rootless-netns auto-heal (problem #1) is intentionally **out of scope** here, but see the note below on why this fix already makes the documented recovery work.

## Root cause

`runStop` deactivates every unit in parallel with mode `"replace"` (`RunParallel` over `podman.StopUnit`). The PHP-FPM unit is `BindsTo` by the per-site worker units, so stopping a worker enqueues a dependency-driven stop of FPM, while the explicit `StopUnit(fpm)` enqueues its own stop with `"replace"`. The two jobs collide and systemd reports the result **`canceled`** for one of them. `dbusUnitOp` surfaced that verbatim as `stop … failed: canceled` and never re-issued the job, so the container was left running.

Because `Restart=always` on the FPM unit (`lerd-php-fpm.container.tmpl`) only fires on an *unexpected* exit, a **completed manual stop** brings the container down for good. The problem was never that the container resists stopping; it is that the stop job was dropped.

## Fix

Re-issue a stop once the competing transaction has settled:

- `dbusUnitOp` retries **only on the `canceled` result**, up to `stopRetryAttempts` (4) with a short proportional backoff. `start` and `restart` keep their single-attempt behaviour and byte-identical error strings.
- The retry policy lives in a DBus-free helper `runUnitOpWithRetry`, unit-tested for: canceled-then-done, give-up-after-max (final result surfaced), no-retry on a hard `failed`, done-first-attempt, transport error not retried, and `maxAttempts` clamping.

A force `podman rm -f` fallback was considered and rejected: while a unit is still active it would fight `Restart=always` (the container would be resurrected), whereas retrying the stop addresses the actual root cause and leaves systemd in a consistent state.

## Why this also unblocks problem #1's recovery

The report notes that `lerd stop && lerd start` does **not** restore container egress after a host network change, because the one surviving container blocks teardown of the rootless network namespace (the slirp4netns uplink is only rebuilt once the last container exits). With stop now completing reliably, no container survives, the namespace is torn down, and `lerd stop && lerd start` rebuilds the uplink on its own, so the manual `podman rm -f` step in the report's workaround is no longer needed. Detecting a host network change and rebuilding the uplink *without* a stop/start is a separate, larger change and is not attempted here.

## Verification

This is Linux-only code (`//go:build linux`). I develop on macOS, so the package was:

- cross-compiled clean: `GOOS=linux go build ./internal/systemd/`, `go vet`, and a full `CGO_ENABLED=0 GOOS=linux go build -tags nogui ./...`;
- **run on real Linux** (amd64 `golang:1.25` container): `go test ./internal/systemd/` passes, including the new `TestRunUnitOpWithRetry` (6 subtests).

I was not able to reproduce the live `lerd stop` cancellation end-to-end on a real rootless-Podman host, so a confirmation from someone who can hit #530 in the wild would be welcome.

Refs #530.


prabowosd:
@geodro when you have a cycle, this is a scoped fix for the second of the two problems in #530 (the `lerd stop ... failed: canceled` / container-left-running half). I deliberately left the slirp4netns uplink auto-heal (problem #1) out of this PR, since it's a larger, platform-specific change, and I'd rather land the stop-reliability fix first: it already turns the documented `lerd stop && lerd start` into a working recovery (a surviving container no longer blocks the rootless-netns teardown). Happy to take the uplink auto-heal on as a follow-up if you'd like it, or to defer to your call on the approach (rebuild on host network change vs. the `pasta` backend the reporter suggested).

One caveat on verification: I develop on macOS, so I cross-compiled the package and ran the `internal/systemd` tests on a real Linux container, but I could **not** reproduce the live `lerd stop` cancellation on actual rootless Podman. The root cause is reasoned from the code (parallel `"replace"` stops of the `BindsTo`-linked worker and FPM units cancelling each other), not from a live repro.

@D90-dev, since you can hit #530 in the wild, could you test this branch on your host and push back if it doesn't hold up? To try it, from a clone of geodro/lerd:

```sh
git fetch origin pull/532/head:pr-532
git checkout pr-532
make build                # or: make install
lerd stop                 # should no longer print: stop lerd-php85-fpm failed: canceled
podman ps                 # no lerd-* container should be left running
```

And the part that matters most for your report: after a host network change (Wi-Fi toggle / suspend-resume), a plain `lerd stop && lerd start` should now restore container egress on its own, **without** the manual `podman rm -f` step from your workaround. If a container still survives the stop, or egress stays dead, please paste the exact `lerd stop` output and `podman ps`, and I'll dig back in.

Thanks both!

geodro:
had a read through this and the reasoning holds up. the canceled result really is the parallel replace stops of the worker and fpm units cancelling each other, and re issuing the stop once the competing transaction settles is the right call. scoping the retry to stop only looks correct too, timeouts return straight away so they never stack, and giving up still surfaces the same canceled error as before so nothing downstream changes.

worth adding one data point on the live repro. i ran lerd stop five times in a row on a host with queue and schedule workers that BindsTo the fpm unit, on podman 5.8.2, and the fpm units stopped cleanly every time, no canceled and no leftover container, so i could not trigger it either. that matches what you saw, and my guess is the race is specific to the older 4.9.3 plus slirp4netns stack in the report rather than something newer podman still hits, which is why the real confirmation belongs with @D90-dev on their box.

no concerns from me on the code. happy to land it once D90 confirms it holds up in the wild, and the uplink auto heal can come as the follow up you suggested.

prabowosd:
Ok just wait if @D90-dev can test on real machine

prabowosd:
@geodro just land this PR we can't push back to @D90-dev still not response. And i found a new bugs on my mac

PR #549, opened 2026-06-18
feat: support macOS in the one-line installer

geodro:
Recent Homebrew versions block third-party taps until you run `brew trust geodro/lerd`, so the tap stopped being a frictionless install path. This makes our own `install.sh` cross-platform so the same curl one-liner works on macOS, the way Linux already installs. All the heavy lifting like Podman Machine, mkcert, the `/etc/resolver` DNS entry, and the launchd plists already lives in `lerd install`, which is cross-platform, so the script only needs to deliver the right binary and hand off.

On the installer side, `detect_os` now branches between linux and darwin, `detect_arch` maps macOS `arm64`, and `download_binary` pulls the `darwin_<arch>` release asset. macOS prerequisites only check for the `podman` CLI and install anything missing through Homebrew without sudo, while the uninstall path boots out the launchd agents (reading each plist's real `com.lerd.*` Label rather than guessing it from the filename) and removes any detached `lerd-*` containers. The bash PATH line is written to `.bash_profile` on macOS since Terminal launches bash as a login shell that reads `.bash_profile` rather than `.bashrc`, and the macOS uninstall avoids a `set -o pipefail` abort when no `lerd-*` containers match the cleanup grep.

The binary mirrors that PATH choice, so `addShellShims` writes to `.bash_profile` on darwin through a new `bashRCPath` helper and keeps `.bashrc` on Linux. `lerd update` now self-updates on macOS too: it downloads the darwin archive for the running architecture and reloads the launchd agents through the shared service manager so the UI and watcher pick up the new binary, while still deferring to `brew upgrade lerd` when the binary lives under a Homebrew Cellar.

The README and the installation guide now lead with the macOS curl installer and keep Homebrew as an alternative with a note about `brew trust`. No new release is needed since the one-liner serves `install.sh` from main and the macOS release binaries already ship.


geodro:
@prabowosd can you test this installer ?

prabowosd:
I will check it soon asap

prabowosd (approved):
Tested the macOS paths on real hardware (Apple Silicon, macOS 15, zsh as login shell, podman + brew present). CI is green and the darwin branches all resolve correctly:

- `detect_os` -> `darwin`, `detect_arch` -> `arm64` (the macOS `arm64` -> release `arm64` mapping works)
- `detect_shell_rc`: zsh -> `~/.zshrc`, bash -> `~/.bash_profile` (the login-shell fix is correct)
- `--check` is read-only and reports prerequisites met; `--help` is correct
- resolved asset name: `lerd_<ver>_darwin_arm64.tar.gz`
- the "already up to date" guard fires when installed == latest and exits before any download/overwrite

A couple of non-blocking notes for follow-up, not merge blockers:

1. `install.sh --uninstall` on its own removes the binary and launchd agents but leaves `/etc/resolver` and the Podman machine behind (by design, since the resolver needs sudo via `lerd uninstall`). The docs cover the two-step order, but a user who runs only the one-liner uninstall ends up with an orphaned resolver entry. Could be worth a hint in the uninstall output pointing at `lerd uninstall` when the binary is still present.

2. For anyone running an ahead-of-release local build at `~/.local/bin/lerd`, re-running the one-liner would overwrite it with the release binary once a newer release exists (`installed_version`'s `x.y.z` regex also treats a git-describe build as the base release, so an equal base is treated as up to date). Edge case, just flagging.

LGTM for the macOS one-liner. Nice that all the heavy lifting stays in `lerd install`.

prabowosd:
@geodro tested the macOS installer on real hardware (Apple Silicon, zsh login shell) and left an approving review above. The darwin paths all check out and CI is green. Two non-blocking notes in the review for follow-up whenever you get to it. Good to merge from my side.

geodro:
Thanks for the thorough test pass on real hardware. I addressed both of your follow up notes in the latest commit.

For the uninstall one, install.sh --uninstall on macOS now checks whether the lerd binary is still present, and when it is it warns that it won't remove the DNS resolver at /etc/resolver/test or the Podman machine, then pauses so you can bail out and run lerd uninstall first, matching the documented two step order. In the non interactive curl pipe case where it can't prompt, it prints the manual removal commands at the end instead, so the resolver can't end up silently orphaned either way.

For the version comparison one, the installer now keeps the full version token instead of just the bare x.y.z, recognises a local development build by its git describe suffix, and asks before replacing it with a release binary. So an ahead of release build at ~/.local/bin/lerd is kept rather than overwritten once a newer release exists. Declining, or running with no tty, keeps the build and exits cleanly, and you can still pull a dev build back in with install.sh --local.

I also added a few bats cases around the new version helpers and documented both behaviours in the installation docs.

prabowosd:
Re-verified both follow ups on the latest commit, they look great.

**Dev build guard:** confirmed `version_is_dev` classifies correctly in isolation: git-describe builds (`1.25.0-6-g7d030096`, `1.24.1-84-gc3f2116`, even `-dirty`) are treated as DEV and prompt before overwrite, while clean releases (`1.25.0`, `12.3`) pass through. `guard_dev_build` runs after the up-to-date check in both `cmd_install` and `cmd_update`, and declining or a missing tty keeps the local build and exits cleanly. Nice touch keeping `installed_version_raw` for the suffix while `installed_version` stays bare for the equality check.

**Uninstall resolver:** the pre-teardown warning + confirm nudges the two-step order while the binary is present, and the closing `/etc/resolver/test` check prints the manual `sudo rm` + `podman machine rm` commands so the curl-pipe (no tty) path can't silently orphan the resolver either. Exactly what was needed.

Both notes fully addressed. My approval stands, good to merge.

PR #551, opened 2026-06-18
Update list of available MariaDB images

Sn0wCrack:
Update to include MariaDB 12 (latest), 12.3 LTS, 11.8 LTS (same as previous 11 (latest) and 11.4 LTS.

This covers all currently active supported MariaDB versions now.


prabowosd:
Verified the tags and labels, all check out:

- `docker.io/library/mariadb:12.3` exists and `12.3` is indeed the newest LTS (released 2026-05-28, EOL 2029-06-30). `11.8` / `11.4` / `10.11` are all current LTS lines too, so the list is accurate.

Two minor notes, non-blocking:

1. Dropping the bare `11` tag means existing `11` users get orphaned (their `lerd-mariadb-11` container and per-tag data dir are no longer referenced, and `11.8` reuses host_port 3411). Might be worth keeping an `11` alias or a note so existing installs don't silently lose access to old data.
2. `12 (latest)` and `12.3 LTS` both resolve to 12.3 right now, but that's fine going forward as `12` rolls ahead while `12.3` stays the pinned LTS.

Content looks good and current overall.

PR #552, opened 2026-06-19
feat(ui): move editors to Monaco and back tinker with phpantom_lsp

geodro:
The dashboard's in-browser editors move from CodeMirror to the Monaco editor across every surface: the tinker REPL, the per-site and global nginx overrides, php.ini and service tuning files, and the site .env editor. Monaco is lazy-loaded so it stays out of the initial dashboard bundle (the always-loaded chunk actually shrinks), and the nginx, ini, and dotenv formats get lightweight Monarch grammars themed to match the rest of the UI. A shared MonacoEditor wrapper keeps the same value/readOnly/onChange contract the old CodeEditor exposed, so the config tabs, save modals, validation, and backup flows are untouched.

Tinker's autocomplete, diagnostics, hover, and signature help now come from phpantom_lsp, a self-contained Rust PHP language server, replacing the hardcoded Laravel and Eloquent completion lists and the php -l check. Because it analyses the real project rather than guessing from static lists, completions resolve actual models, relationships, scopes, and vendor classes. lerd manages the binary on the host alongside fnm and mkcert, fetching it at install time on both Linux and macOS and lazily on first use for existing installs. The browser talks to it over a new /api/lsp/php WebSocket that spawns one server per connection rooted at the site or worktree path and bridges the browser's one-message-per-frame traffic to the server's stdio Content-Length framing. Tinker buffers are headerless PHP, so the bridge presents the document to the server with a synthetic leading <?php line and offsets positions accordingly, keeping the editor headerless while the server sees valid PHP.

The CodeMirror dependency is dropped entirely, along with the now-unused tinker symbol and lint endpoints and their backing code. If the language server can't be reached the editor still works and code still runs; only the live intelligence is unavailable, surfaced by a small status hint.


AJenbo:
cool 😀

PR #588, opened 2026-06-22
feat: host (system) database mode (MySQL/MariaDB/Postgres) via `db.external`, with port-ownership guard

iabduul7:
## What

Adds an opt-in **host (system) database mode**: a project sets `db.external: true` in `.lerd.yaml` to point its containerized PHP at a **host-installed database** instead of lerd's container — Herd-style. Per-site, with a global default. Generalized across **MySQL, MariaDB, and PostgreSQL** through a single per-engine descriptor, plus a **port-ownership guard** so a host server and lerd's container DB coexist on one machine without fighting for the default port.

## Why

Some setups keep certain databases on the host's system server while everything else runs in lerd. Today both fight for the engine's default port (`3306` / `5432`). This lets them coexist: container projects keep using `lerd-mysql` / `lerd-postgres` over the bridge; host-backed projects reach the system server over its unix socket (Linux) / TCP (macOS); and lerd automatically moves its container's published port out of the way when a host server already owns the default.

## How it works

- **Per-engine descriptor (`HostBackendSpec`)** — one code path now drives host-mode probing, env emission, the FPM socket mount, the dashboard, and the port override for every host-capable engine. Each spec captures the default port, the Linux socket (a **file** for MySQL, a **directory** for Postgres — libpq forms `.s.PGSQL.<port>` inside it), the framework env contract, the container name, and server **install markers**. Adding an engine is a single entry.
- **Engine (`db.external`)** — `lerd env` emits the host connection vars and marks the DB externally-managed so the container isn't started for that site. **Linux**: connects over the host unix socket (bind-mounted into FPM only, existence-gated so a down host server can't break FPM start). **macOS**: connects over TCP to `host.containers.internal` (gvproxy), since unix sockets don't traverse the podman-machine boundary.
- **Port-ownership guard** — at quadlet-**write** time, if a containerized DB would publish its default port but a **host server already owns it**, lerd auto-shifts the container to the next free port, persists the choice, and prints a note. The check is **liveness-independent** (keyed off a non-empty install marker, so a *stopped* host server still forces the shift) and bind-tests dual-stack (`127.0.0.1` + `[::1]`), so the port baked into the `.container` file — which systemd's boot autostart reads with no lerd process running — can never collide and take a host DB down. Manual override stays: `lerd service port <service> [port] [--reset]`.
- **Connection URL** now reflects the published port, so the dashboard's "open connection URL" targets lerd's container — not a coexisting host server sitting on the default port.
- **Dashboard** — engine-neutral per-site backend toggle + global default on each DB service page, a port-aware host-status probe (port-owner attribution: host / lerd / none), a "coexisting with system …" callout, and the published-port control. Loopback-gated; the host option is disabled on confirmed LAN dashboards. All copy parametrized by engine across all 14 locales.
- **Lifecycle** — a host-backed site no longer counts as using the container service; the container auto-stops when truly unused (honoring pin / manual-start) and starts when a container site needs it. `lerd service port` / `expose` on a **removed** service saves the preference without resurrecting its quadlet (so a removed DB can't reappear on boot and squat a host-owned port). Toggling `db.external` off clears the stale `DB_SOCKET` host mode wrote.

## Safety

- Host mode is **loopback-only** — it depends on a host-local socket / `127.0.0.1`, and the host option is rejected from a LAN-exposed dashboard.
- In-container `mysql` / `mariadb` admin runs over **TCP loopback** (`-h127.0.0.1`), independent of the container's socket path, and SQL identifiers / literals are escaped at the sink.

## Scope / notes

- `migrate=copy` (dump → import on a backend switch) is intentionally rejected for now — this points the connection only, it does not move data.
- Postgres host mode authenticates over the socket; a containerized client can't use `peer` auth, so a `local … scram-sha-256` line in `pg_hba.conf` is needed (lerd surfaces a note — edit it preserving `postgres:postgres` ownership; a root `sed -i` leaves it root-owned and the cluster won't restart).
- Default published ports stay at the engine default, so existing installs are untouched until a host server is detected or the user moves the port.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean; `go test ./...` green — incl. new tests: the port guard (bindable / first-free / install-marker liveness-independence), `ServicePublishedPort`, `WithURLPort`, `MySQLAdminCmd` TCP loopback, the port-owner attribution matrix, host-backend `SitesUsingService` exclusion, `applyHostDBExternalEnv` for the Linux socket **and** macOS TCP, the descriptor, and the stale-socket clear.
- Frontend: `npm run check` (svelte-check + locale parity) clean.
- **Validated end-to-end on a real Linux machine**: started `lerd-postgres` alongside a host PostgreSQL — it auto-shifted to a free port (skipping another busy port), both reachable, loopback-only; proved which server a Laravel app resolves to in host vs container mode (socket vs TCP, via distinctly-named marker DBs); and visually verified the dashboard renders correctly for both MySQL and Postgres.

Rebased on current `main`.


geodro:
Thanks for this, the host database mode is a genuinely useful direction and the engine plumbing is solid. Before we go deep on review there's a design question I want to settle first.

Right now this is really MySQL only. The lerd-mysql service already backs both MySQL and MariaDB so those aren't two separate cases, and Postgres is explicitly out of scope, which means the feature ships host mode for exactly one database. Pointing a site at a host installed server instead of the container is a general capability though, there's nothing MySQL specific about the idea, and Postgres is the first thing people will reach for next.

The trouble is the implementation is shaped around MySQL rather than being generic. The probe attributes the 3306 owner, the port override is a mysql preset, the socket path is the MySQL one, and the env emission is keyed to the MySQL connection vars. So this isn't a case where Postgres slots in later for free, adding it would mean reworking those paths.

Could you rework the design so the external backend and the configurable published port are per service capabilities rather than MySQL specific code, so at least all databases share the same mechanism. You can wire up MySQL and Postgres to start, but I'd like the foundation to generalise so we're not bolting on a parallel implementation for every engine. Happy to talk through where the service abstraction should live if that helps.

iabduul7:
Glad to hear and thank you for reviewing @geodro. I will work on it and update the PR

prabowosd:
Chiming in as someone running lerd on macOS with a lot of databases consolidated into `lerd-mysql`. The coexistence problem this solves is real, so +1 on the direction. Two things from the macOS / multi-DB angle worth baking in while the design gets generalized:

**1. Host-grant matching on the macOS TCP path.** When the app connects through `host.containers.internal` (gvproxy), the host MySQL sees the connection from the VM/gateway source IP, not loopback. A user granted only `'app'@'localhost'` (or `@'127.0.0.1'`) will hit `Access denied`, and the server has to listen on an interface gvproxy can actually reach (a `bind-address=127.0.0.1`-only setup may not be enough). Same "Linux socket works, macOS TCP surprises you" shape as the debug-dump bridge. A one-line hint in the `lerd env` output or the dashboard callout ("on macOS host mode, grant on `'%'` and bind beyond loopback") would save people a confusing first connection.

**2. Moving `lerd-mysql`'s published port.** Container projects are unaffected since the internal port stays 3306, but host-side tooling pinned to the *published* 3306 (GUI clients, socket forwarders, scripts) has to follow the move. Worth surfacing in the port-change confirmation.

On the generalization ask: agree, and `PublishedPort` already living on `ServiceConfig` is a good anchor. If the host-mode bits (default port, socket path, env var names, family predicate, probe owner) become a per-engine descriptor, MySQL and Postgres are just two table entries and the env emission / probe / port override stop being MySQL-shaped. Postgres has the identical 5432 conflict, so it wants the exact same mechanism.

Happy to test the macOS path on a real M-series machine once the rework lands.

iabduul7:
Pushed a substantial update and rebased onto current `main`. What changed since the original host-MySQL scope:

**Generalized beyond MySQL.** Host mode is now driven by a single per-engine descriptor (`HostBackendSpec`), so MySQL, MariaDB, and **PostgreSQL** share one path for probing, env emission, the FPM socket mount, the dashboard, and the port override. Postgres connects over the socket **directory** (libpq forms `.s.PGSQL.<port>` inside it); MySQL over the socket **file** — the descriptor captures the difference. Adding an engine is one entry.

**Port-ownership guard.** At quadlet-write time, if a containerized DB would publish its default port but a host server already owns it, lerd auto-shifts the container to the next free port and persists it. It's liveness-independent (keyed off a non-empty install marker, so a *stopped* host server still forces the shift) and bind-tests dual-stack — so the port baked into the `.container` file, which boot autostart reads with no lerd process running, can never collide and take a host DB down.

**Connection URL** now reflects the published port (it was advertising the engine default — which, in the guard's scenario, is exactly the port a coexisting host server sits on).

**Lifecycle hardening.** `lerd service port` / `expose` on a *removed* service no longer resurrects its quadlet; toggling `db.external` off clears the stale `DB_SOCKET` host mode wrote.

**Dashboard** is now engine-neutral — per-site backend toggle, host-status probe with port-owner attribution, a "coexisting with system …" callout, and the published-port control — with all copy parametrized by engine across the 14 locales.

**Validated end-to-end on Linux:** brought up `lerd-postgres` alongside a host PostgreSQL — it auto-shifted to a free port (skipping another busy one), both reachable and loopback-only; confirmed which server a Laravel app resolves to in host vs container mode (unix socket vs TCP, via distinctly-named marker DBs); and visually verified the dashboard renders correctly for both MySQL and Postgres.

Rebased on current `main` (reconciled with the recent SQL-escaping hardening — in-container admin keeps both the TCP-loopback transport and the identifier/literal escaping). Full gate green: `go build/vet/test ./...`, `npm run check`.

prabowosd:
i will checked first thanks @iabduul7

iabduul7:
Thanks @prabowosd — and good timing: the generalization you described is exactly what just landed. Host mode now runs off a per-engine descriptor (`HostBackendSpec`: default port, socket — a *file* for MySQL, a *directory* for Postgres — env contract, family predicate, container/probe owner), so MySQL, MariaDB and Postgres are table entries and the env emission / probe / port override are no longer MySQL-shaped. Postgres' 5432 conflict gets the identical mechanism — a port-ownership guard auto-shifts `lerd-postgres` off the default when a host server owns it (validated live: host PG on `:5432`, lerd-postgres auto-placed on a free port, both reachable, loopback-only).

On your two macOS points — both are real and currently unaddressed, so I'll fold them in:

**1. macOS TCP grant/bind.** Agreed. The gvproxy path means the host server sees a non-loopback source, so a `@'localhost'` / `@'127.0.0.1'`-only grant gives `Access denied`, and a `bind-address=127.0.0.1`-only server isn't reachable at all. I'll add an engine-aware note to the macOS branch of `lerd env` (and the dashboard host-setup callout): for MySQL, grant on `'%'` (or the gateway) and listen beyond loopback; for Postgres, a `host … scram-sha-256` `pg_hba` line plus `listen_addresses`. It'll sit right where the Linux Postgres peer-auth note already lives, so the "socket works, TCP surprises you" gap is closed on both engines.

**2. Published-port move.** Fair — the confirmation notes the container-internal port is unchanged but doesn't flag that host-side tooling pinned to the *published* port (GUI clients, forwarders, scripts) has to follow. I'll surface that in both the manual `lerd service port` confirmation and the automatic guard's "publishing on `:<n>`" note.

I'd love to take you up on the M-series test once these land — I've validated the Linux socket + coexistence end-to-end, but the macOS gvproxy grant/bind path is exactly the bit I can't exercise here.

prabowosd (changes requested):
Backend-focused review (UI/i18n left to @geodro). Solid design overall: the engine-family `HostBackendSpec` descriptor is clean, the install-marker liveness-independence (checking a populated cluster dir rather than a live socket) is the right fix for the boot-ordering race, and the env matrix is well table-tested. A few real gaps worth fixing before merge.

## MAJOR

**1. `lerdReservedPorts()` ignores preset `Port`, only reserves `PublishedPort` + `ExtraPorts` (`internal/serviceops/serviceops.go`)**
The auto-picker excludes user overrides but not each service's preset-default `Port`. A *stopped* lerd service sitting on its default port (e.g. mariadb preset 3307) is invisible to the reserved set, `portBindable` returns true (nothing is listening), and the guard can hand that port to MySQL. At boot both units race to bind it. This directly undermines the stated goal of "never auto-picks a port another lerd service will bind". One-line fix: also `if svc.Port > 0 { reserved[svc.Port] = true }` in the loop.

**2. `databaseExists` keeps an unescaped `schema_name='%s'` (`internal/cli/db.go:527`)**
The PR swaps the transport to `MySQLAdminCmd` (good, fixes TCP loopback) but leaves the query interpolating a bare `name`, while `provision.go` escapes the same query. A DB name containing a quote (e.g. from a branch like `it's-a-test`) breaks the query or injects into a root-level query inside the container. Exploitability is low (name comes from local project config, not remote input), but since the PR already touches this line, reuse the same escaping (`EscapeSQLLiteral`).

**3. Port-guard swallows the `SaveGlobal` error (`internal/serviceops/serviceops.go`)**
In the guard's save-then-commit block, if `SaveGlobal` fails the error is discarded via `== nil`, `publishedPort` stays 0, and the quadlet is written with the default port, colliding with the host server, silently. This is the "return on a real failure, don't WARN-and-continue" pattern: surface the error instead of swallowing it.

## MINOR

- **`TCPPortBusy` (dial) is weaker than the guard's `portBindable` (bind + dual-stack v4/v6).** The CLI/dashboard pre-flight can accept a port already bound only on `::1`; the quadlet guard then auto-shifts it unexpectedly. Align the pre-flight to `!portBindable(port)`.
- **`firstFreeHostPort` returns 0 silently** when no port is free, and the guard falls through to the default port with no diagnostic. Print a warning on the `free == 0` path.
- **Postgres socket probe hardcodes `spec.DefaultPort`** for `.s.PGSQL.<port>`, so a host cluster on a non-default port (5433) is missed and the guard never sees it. Glob `.s.PGSQL.*` via `ReadDir` instead.
- **`ReconcileDBServiceLifecycle` covers only mysql/mariadb, not postgres.** After a bulk switch to host Postgres, `lerd-postgres` keeps running indefinitely. Add `"postgres"` (or derive from the specs).
- **`apply_all` in `handleSettingsDefaultBackend` does not skip paused/ignored sites**, unlike `SitesUsingService`. Add `if s.Ignored || s.Paused { continue }`.
- **IPv6 bind-test TOCTOU in `portBindable`** (v4 listener closed before binding v6). Low probability; hold both open if convenient.

## NIT

- `actual` port re-read: a map miss on `cfg2.Services[name]` silently yields 0. Use the comma-ok form.
- No `CollectPortChecks` test for the `PublishedPort > 0` branch.

## What is well done

Install-marker liveness-independence, `SetPrimaryHostPort` being pure + immutable (slice copied), `socketLive`'s two-signal stale-vs-live distinction, the validate-before-disk ordering in the guard, and the macOS TCP path via `host.containers.internal` with the socket-mount short-circuit, all correct and covered by tests.

---

Separate process note, not blocking: this PR is +3600 across 52 files in one shot, which makes it hard to review thoroughly and risky to land. Going forward please split work like this into a stack of smaller PRs (e.g. descriptor + env emission, port-ownership guard, lifecycle reconcile, dashboard), the way we do on our side. Each one is easier to reason about, test in isolation, and revert if needed.

prabowosd:
Nice, the per-engine descriptor is exactly the shape this needed, and good to hear the host-PG coexistence checks out live on Linux.

Happy to take the M-series test once the two macOS notes land (the gvproxy grant/bind hint and the published-port move warning). That gvproxy path is the bit you can't exercise from Linux, so I'll bring up a host MySQL and a host Postgres on the Mac, run a site through host mode for each, and confirm the grant/bind guidance actually gets someone connected on the first try, plus that the dashboard callout reads right on macOS.

Separately, I left a backend code review just now with a few correctness items worth folding into the same pass, the headline ones being the `lerdReservedPorts` preset-port gap (a stopped lerd DB on its default port can still get auto-picked) and a swallowed `SaveGlobal` error in the port guard. Details are in the review.

iabduul7:
Thanks @prabowosd — this is a genuinely useful pass, and all three MAJORs are real. Agreed on the whole list; here's how I'll fold them in.

**MAJOR**

1. **`lerdReservedPorts()` ignoring preset `Port`.** You're right, this is a hole in the guarantee — a *stopped* service on its preset default is invisible to `portBindable` (nothing's listening), so the auto-picker can hand it out and both units race at boot. I'll reserve `svc.Port` alongside `PublishedPort`/`ExtraPorts` in the loop, and add a `firstFreeHostPort` test that proves a stopped preset-port service is skipped.

2. **`databaseExists` unescaped `schema_name='%s'`.** Good catch — the transport swap touched the line but left the literal bare while `provision.go` escapes the identical query. I'll route it through the same escaping helper so the two are consistent (and exercise it with a quote-bearing name).

3. **Swallowed `SaveGlobal` error in the guard.** Agreed this is the wrong failure mode: if persistence fails we currently fall through and write the quadlet on the contested default port, silently. I'll surface the error and *not* emit a quadlet on the default port when the shift couldn't be persisted — fail closed rather than collide.

**MINOR** — all fair, taking them all:
- Align the CLI/dashboard pre-flight to `!portBindable(port)` so a `::1`-only bind isn't accepted-then-shifted.
- Warn on the `firstFreeHostPort` `free == 0` fall-through instead of falling back silently.
- Glob `.s.PGSQL.*` via `ReadDir` so a host cluster on a non-default port (5433) is actually probed, instead of hardcoding `DefaultPort`.
- Add `postgres` to `ReconcileDBServiceLifecycle` (or derive the set from the specs) so `lerd-postgres` doesn't linger after a bulk switch to host Postgres.
- Skip `Ignored`/`Paused` sites in `apply_all`, matching `SitesUsingService`.
- Hold both listeners open for the `portBindable` v4/v6 TOCTOU.

**NIT** — comma-ok on the `actual` re-read, and a `CollectPortChecks` test for the `PublishedPort > 0` branch.

**macOS notes (the two you're waiting on):** both are done. `lerd env` now prints an engine-aware gvproxy grant/bind hint on the macOS TCP path (MySQL: bind beyond 127.0.0.1 + grant on `'%'`; Postgres: `listen_addresses` + a `host … scram-sha-256` `pg_hba` line), the dashboard host-setup callout shows the same note when the probe reports TCP, and the published-port move now warns about repointing host clients on the auto-shift path too (the manual `lerd service port` path already did). They'll be in the next push together with the fixes above, so the M-series run exercises everything at once.

**On the PR size / stacking:** completely fair, and I'd rather work that way too — this one ballooned because the engine generalization, the port guard, the lifecycle reconcile and the dashboard all landed together. Since it's already open and mid-review I'll fold these corrections into the existing branch as focused, self-contained commits; for the next chunk of host-DB work I'll stack it as smaller PRs (descriptor+env, guard, lifecycle, dashboard) the way you described.

Really appreciate the M-series offer — I'll ping you once this is pushed so you can bring up host MySQL + Postgres and confirm the grant/bind guidance actually gets someone connected on the first try.

geodro:
I just got the chance to check this in detail. Before we take it further I want to make sure we are closing a real gap, because I think most of what db.external does is already reachable today with the per site override file. If you drop a .env.lerd_override next to your .env with a few lines, lerd stops managing that site's database, leaves your connection values alone, and your containerized PHP reaches the host server over the host.containers.internal alias that lerd already writes and keeps fresh:

```
LERD_EXTERNAL_SERVICES=mysql
DB_HOST=host.containers.internal
DB_PORT=3306
DB_USERNAME=youruser
DB_PASSWORD=yourpass
```

I ran this end to end on Linux, a containerized PHP app connected straight to a host MySQL through that alias and lerd never started its own mysql container or rewrote the env values. So the host reach and the do not manage my database parts both look covered already, which makes me think the missing piece is really that this recipe is not documented anywhere, more than a new code path. I am happy to add a docs page for it.

So my honest question, what does this give a user that the override file does not? If there is a concrete case the override cannot cover I would love to understand it, and if it is mainly about discoverability and having it committed in .lerd.yaml then I would lean toward the docs page and keeping the surface area small, especially with the Postgres pg_hba step that has to be done by hand. Keen to hear your thinking.

geodro:
see https://lerd.sh/features/env-setup#personal-overrides-env-lerd-override

iabduul7:
Thanks for taking the time to dig into this — it's a fair challenge, and you're largely right. The two core things, reaching a host server and lerd not managing that site's DB, are already covered by `.env.lerd_override` + `LERD_EXTERNAL_SERVICES`. A docs page for that recipe would genuinely help, and I'm happy to drop the "discoverable / lives in `.lerd.yaml`" half of this onto it.

There's one thing the override can't do, though, and it's the part I'd argue is worth keeping: **a host server and lerd's own container DB coexisting on the same machine.**

The override points a *site* at the host server, but `lerd-mysql` — still serving your other sites — publishes `127.0.0.1:3306`, the same port the host MySQL sits on, so they fight for it. And the failure lands exactly where you can't fix it by hand: the `.container` quadlet bakes in `:3306`, so at boot systemd autostarts it with no lerd process running to notice the host server and step aside — it just races to bind 3306, and whoever loses goes down. So "one host-backed site alongside other container-backed sites" isn't reachable through the override, and even a manual port pick isn't boot-safe on its own.

That's what the port-ownership guard solves: at quadlet-write time, if a host server already owns the engine's default port — detected liveness-independently via the install marker, so a *stopped* server still counts — lerd shifts its container to the next free port and bakes that into the quadlet, so the boot autostart can't collide.

So my suggestion: take you up on the docs page for the override recipe, drop the committed `db.external` surface, and let me **split out just the port-ownership guard as a small standalone PR** (which also matches @prabowosd's ask to stack these into smaller units). That's the piece that closes a real gap; the rest is convenience the override already gives. Does that split sound right to you?

geodro:
That split is right, let's do it. Drop the committed db.external surface, port ownership guard as its own small PR, and I'll take the docs page for the override recipe since it's a small env-setup edit anyway.

One thing for that PR, keep the manual lerd service port command in it alongside the automatic guard, it is the user facing way to move the published port and useful on its own for anyone with a port clash, and the guard builds on the same machinery.

On the dashboard, leave the service page as it is rather than adding inline controls. If you do want a UI for the port, put it behind a cog in the top right that opens a small modal to set the published port with validation, so it stays out of the way.

One thing to watch on a port change, container sites are fine since they reach the service over the podman network on its internal port, but host-proxy sites connect over the published host port, so those are the only ones whose .env needs to follow the new port.

Appreciate you being open to slimming it down, good outcome.

iabduul7:
Closing this in favor of #633, which carves out just the port-ownership guard (plus the `lerd service port` command and the host-proxy port-follow) as we discussed.

#633 drops the committed `db.external` surface and all the dashboard/UI, so it's a much smaller, focused change. It also folds in @prabowosd's actionable review items from here. The override recipe keeps its docs page.

Happy to reopen or iterate if anything looks off in the slimmer version.

PR #617, opened 2026-06-25
feat(host-proxy): bind dev server to all interfaces so the proxy container can reach it

prabowosd:
## Problem

nginx runs inside a container and reaches host-proxy dev servers over the host **gateway IP**, never loopback. A dev server bound to `127.0.0.1` (the common dev default) is unreachable through the lerd proxy.

How it surfaces depends on the framework:
- generic Node server on loopback → connection error / 502 through the proxy
- **Nuxt** → the main app binds loopback but Vite's HMR socket binds all interfaces on the same port, so the container hits the HMR socket, which answers every plain HTTP request with `426 Upgrade Required`. In the browser this reads as a misleading **"Update required"** page even though the app is perfectly healthy.

Repro: link a Nuxt 4 dev app as a host-proxy site, open the domain → `426`. Direct `curl localhost:<port>` → `200`. The only difference is loopback vs the gateway interface.

## Fix

Mirror the existing `PORT` injection. `buildHostProxyCommandPort` now prefixes the supervised dev command with `HOST=0.0.0.0` alongside the port, so Nuxt, NestJS, and most Node servers bind all interfaces out of the box. The bind env var is configurable via a new `host_env_key` field (default `HOST`, mirroring `port_env_key`) for servers that use a different name, e.g. `HOSTNAME` for a Next.js standalone server.

A `HOST` the user sets later in their own command still wins, because `env` evaluates left to right.

The raw **Vite CLI** ignores the `HOST` env (it reads `--host`), so the `lerd init` Vite reminder and the host-proxy docs are extended to tell those users to add `--host` explicitly.

## Changes
- `internal/config/project.go` — new `host_env_key` field on `ProxyConfig`.
- `internal/cli/hostproxy.go` — `hostProxyHostEnvKey` helper + inject `HOST=0.0.0.0` in `buildHostProxyCommandPort`.
- `internal/cli/init.go` — extend the Vite reminder with the `--host` note.
- `docs/usage/host-proxy.md` — new **Binding** section, `host_env_key` table row, Vite `--host` note.
- `internal/cli/hostproxy_test.go` — updated assertions + custom-env-keys case.

## Test
`go test ./internal/cli/ ./internal/config/` green; `go vet` and `gofmt` clean. Verified end to end against a real Nuxt 4 host-proxy site: domain went from `426` to `200` once the dev server bound all interfaces.

### DIFF
--- a/docs/usage/host-proxy.md
+++ b/docs/usage/host-proxy.md
@@ -49,6 +49,7 @@ The `proxy` section in `.lerd.yaml` is mutually exclusive with `container:` (a s
 | `port` | yes | | Port the dev server listens on. lerd injects this and proxies to it. |
 | `ssl` | no | `false` | Set to `true` if the dev server serves HTTPS on its port (nginx proxies via `https://` with `proxy_ssl_verify off`). |
 | `port_env_key` | no | `PORT` | Environment variable the port is injected as. Most servers read `PORT`; set this if yours uses a different name. |
+| `host_env_key` | no | `HOST` | Environment variable the bind address (`0.0.0.0`) is injected as. Most servers read `HOST`; set this if yours uses a different name, e.g. `HOSTNAME` for a Next.js standalone server. |
 
 ## Ports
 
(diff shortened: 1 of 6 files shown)


geodro (changes requested):
Thanks for the PR, nicely put together with the tests, docs, and the writeup.

One thing I'd like to sort out before merging. Right now the HOST injection is unconditional with no way to turn it off. That's fine for the common case where HOST is the bind address, but a dev server that reads HOST for something else, or a project that already sets HOST in its .env, gets it silently forced to 0.0.0.0, and because dotenv won't override a variable that's already in the process environment the app ends up seeing 0.0.0.0 instead of its own value. Pointing host_env_key at a different variable helps, and setting HOST yourself later in the command still wins, but there's no way to suppress the injection entirely, so it's a behavior change existing host proxy sites can't opt out of.

Could you add an escape hatch for that, maybe a bind flag that defaults to true so people can set it false, or some other opt out you prefer. Once that's in I'm happy to merge.

prabowosd:
Good catch, addressed in 63c2a94.

Added a `bind` field to `ProxyConfig` (`*bool`, defaults to `true` so existing sites are unaffected). When set to `false`, `hostProxyShouldBind` short-circuits and the command keeps only the port injection, so a server that reads `HOST` for its own purpose or already sets it in `.env` can opt out entirely:

```yaml
proxy:
  command: npm run dev
  port: 3000
  bind: false
```

Two new `buildHostProxyCommand` tests cover it (`bind: false` opts out, `bind: true` is identical to the default), and the docs get a `bind` table row plus a note in the Binding section calling out the dotenv-already-in-env case you described. `go test ./internal/cli/ ./internal/config/` green, `go vet` and `gofmt` clean.

geodro:
Thanks for turning the opt out around so fast. I pushed one small tweak straight onto the branch so we did not have to bounce it back again, I renamed the field from bind to inject_host. On its own bind: false reads a little like turn the dev server off entirely, but the field is really about whether lerd injects the HOST env, so inject_host: false says exactly what it does and it sits nicely next to port_env_key and host_env_key. Behaviour is unchanged, it still defaults to true so existing sites are untouched, and I updated the tests and docs to match. Green locally, good to merge from my side once CI passes.

prabowosd:
All yours, LGTM

PR #618, opened 2026-06-25
fix(macos): keep published ports when lan.exposed is true

prabowosd:
## Problem

On macOS with `lan.exposed: true`, all core containers come up with **no published ports**. `lerd doctor` reports everything green (nginx running, ports 80/443, DNS ok), but `.test` domains are unreachable: the host has nothing listening on 80/443.

Inspecting `lerd-nginx` confirmed it launched with zero `-p` flags (`HostConfig.PortBindings: {}`), only the internal `80/tcp` EXPOSE. The generated `~/Library/LaunchAgents/lerd-nginx.plist` had no publish entries in `ProgramArguments`.

## Root cause

On macOS, `WriteContainerUnit` (`internal/services/launchd_darwin.go`) derives the `podman run` args from the quadlet and runs `stripIPv6PublishPorts` so gvproxy never binds an IPv4 and an IPv6 address on the same port.

With `lan.exposed: true`:

1. `BindForLAN(content, true)` leaves bare `PublishPort=80:80` as-is.
2. `PairIPv6Binds` collapses the bare bind into the bind-all **`[::]:80:80`** form only (no IPv4 partner line is kept for bare/`0.0.0.0` binds).
3. `stripIPv6PublishPorts` then dropped **every** bracketed line, removing the only publish lines that existed.

Result: `podman run` with no `-p`. In loopback mode (`lan.exposed: false`) the bug is hidden because `PairIPv6Binds` keeps the IPv4 `127.0.0.1:` line alongside the `[::1]:` one, so the IPv4 publish survives the strip.

## Fix

`stripIPv6PublishPorts` now rewrites bind-all `[::]:PORT` lines to `0.0.0.0:PORT` instead of dropping them; only loopback `[::1]:` duplicates (which always retain a surviving `127.0.0.1:` partner) are removed. `stripPrivilegedIPBind` then collapses `0.0.0.0:80:80` to `80:80` as before, so privileged-port output is unchanged.

## Tests

- `TestStripIPv6PublishPorts`: unit coverage for `[::]:` rewrite, `[::1]:` drop, and IPv4 / non-publish pass-through.
- `TestLANExposedPublishPortsSurvive`: end-to-end regression running the full `BindForLAN(true)` -> `PairIPv6Binds` -> `stripIPv6PublishPorts` -> `containerToPodmanArgs` chain and asserting `-p 80:80` and `-p 443:443` survive. Fails on the old logic.

Full `internal/services` package is green.

### DIFF
--- a/internal/services/launchd_darwin.go
+++ b/internal/services/launchd_darwin.go
@@ -351,18 +351,31 @@ func stripPrivilegedIPBind(port string) string {
 	return port
 }
 
-// stripIPv6PublishPorts removes PublishPort= lines that start with a bracketed
-// IPv6 address (e.g. "[::1]:3306:3306"). gvproxy cannot bind both an IPv4 and
-// an IPv6 loopback address on the same port simultaneously.
+// stripIPv6PublishPorts normalises bracketed IPv6 PublishPort= lines for gvproxy,
+// which cannot bind both an IPv4 and an IPv6 address on the same port at once.
+//
+// Loopback IPv6 lines ("[::1]:3306:3306") are dropped: PairIPv6Binds always pairs
+// them with an IPv4 "127.0.0.1:" line that survives, so the published port lives on.
+//
+// Bind-all IPv6 lines ("[::]:80:80") have no IPv4 partner — PairIPv6Binds collapses a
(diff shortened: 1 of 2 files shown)


geodro (approved):
Thanks @prabowosd !

prabowosd:
Sure man @geodro !

PR #619, opened 2026-06-25
fix(macos): skip the LAN DNS forwarder where lerd-dns binds the LAN port

prabowosd:
## Problem

On macOS, `lerd lan expose` breaks `.test` DNS resolution. It reports success, but afterwards `lerd-dns` is dead and nothing resolves:

- `lerd-dns` (launchd `com.lerd.lerd-dns`) exits with code 2 (`address in use`).
- `lerd-dns-forwarder` is left bound to `lanIP:5300`, forwarding to a dead `127.0.0.1:5300`.
- `dig @127.0.0.1 -p 5300 <site>.test` times out; the browser can't resolve any `.test` site.

A re-`expose` (the common case) previously failed even earlier, at the preflight, with a spurious `lanIP:5300 is already in use` against lerd's **own** `lerd-dns`.

## Root cause

`EnableLANExposure` unconditionally installs and starts `lerd-dns-forwarder`. That forwarder is the **Linux rootless-pasta workaround** (`installDNSForwarderUnit` documents it as such): on Linux the `lerd-dns` container can't bind the host LAN address, so a host-side forwarder bridges `lanIP:5300 → 127.0.0.1:5300`.

On macOS `lerd-dns` is a host dnsmasq that already binds **every** interface, including `lanIP`. Installing the forwarder makes a second process grab `lanIP:5300`, the `lerd-dns` restart then races it and exits `address in use` (exit 2), and the forwarder forwards to a now-dead loopback. DNS is fully broken.

The preflight check (`preflightForwarderPort`) only recognised lerd's own daemon via the `lerd-dns-forwarder` unit. On macOS that unit is `.disabled` and `lerd-dns` itself holds the port, so the preflight read it as a foreign conflict, which accidentally masked the deeper bug by aborting expose.

## Fix

Gate the forwarder install/start on a new `lerdDNSBindsLANPort` seam (`true` on darwin). Where `lerd-dns` binds the LAN port directly, the forwarder is skipped as a no-op (lerd-dns binding all interfaces *is* the LAN exposure). The Linux model is unchanged: there the forwarder is still installed and the existing preflight still guards it.

Extracts the forwarder steps into `ensureLANForwarder` (the gate) + `installAndStartForwarder` (the impl, behind the `installLANForwarderFn` seam).

## Tests

- `TestEnsureLANForwarder_DarwinSkipsForwarderInstall`: macOS model → forwarder not installed, no-op.
- `TestEnsureLANForwarder_NonDarwinInstallsForwarder`: Linux model → forwarder installed with the right lanIP.
- Existing `preflightForwarderPort` / `forwarderPortFree` tests unchanged and green.

Verified end to end on macOS: `lerd lan expose` now reports `lerd-dns binds the LAN address directly; no forwarder needed`, `lerd-dns` stays up, `.test` resolves, and the site is reachable from a LAN device.

`go test ./internal/cli/` green, `go vet` and `gofmt` clean.

---

Note: this supersedes the earlier preflight-only version of this PR (which let expose proceed but did not prevent the forwarder double-bind). Force-pushed.

### DIFF
--- a/internal/cli/dns.go
+++ b/internal/cli/dns.go
@@ -104,23 +104,50 @@ func EnableLANExposure(progress LANProgressFunc) (lanIP string, err error) {
 			return "", err
 		}
 
-		if err := preflightForwarderPort(lanIP, emit); err != nil {
+		if err := ensureLANForwarder(lanIP, emit); err != nil {
 			return "", err
-		}
-
-		emit("Installing lerd-dns-forwarder.service")
-		if err := installDNSForwarderUnit(lanIP); err != nil {
-			return "", fmt.Errorf("installing dns forwarder: %w", err)
-		}
-
-		emit("Starting lerd-dns-forwarder")
-		if err := reloadAndRestartUnit("lerd-dns-forwarder"); err != nil {
-			return "", fmt.Errorf("starting dns forwarder: %w", err)
 		}
 	}
 
 	emit("Done — lerd is reachable on " + lanIP)
 	return lanIP, nil
+}
+
+// ensureLANForwarder installs the host-side LAN DNS forwarder when the platform
+// needs it. The forwarder is the Linux rootless-pasta workaround: there the
+// lerd-dns container cannot bind the host LAN address, so a forwarder bridges
+// lanIP:5300 → 127.0.0.1:5300. On macOS lerd-dns is a host dnsmasq that already
+// binds all interfaces (including lanIP), so installing a forwarder would
+// double-bind lanIP:5300 and crash lerd-dns; there this is a no-op.
+func ensureLANForwarder(lanIP string, emit func(string)) error {
+	if lerdDNSBindsLANPort {
+		if emit != nil {
+			emit("lerd-dns binds the LAN address directly; no forwarder needed")
+		}
+		return nil
+	}
+	return installLANForwarderFn(lanIP, emit)
+}
+
+// installAndStartForwarder runs the preflight, installs the lerd-dns-forwarder
+// unit, and starts it. Default implementation behind installLANForwarderFn.
+func installAndStartForwarder(lanIP string, emit func(string)) error {
+	if err := preflightForwarderPort(lanIP, emit); err != nil {
+		return err
+	}
+	if emit != nil {
+		emit("Installing lerd-dns-forwarder.service")
+	}
+	if err := installDNSForwarderUnit(lanIP); err != nil {
+		return fmt.Errorf("installing dns forwarder: %w", err)
+	}
+	if emit != nil {
+		emit("Starting lerd-dns-forwarder")
+	}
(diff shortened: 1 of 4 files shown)


geodro:
Nice fix, the seam approach reads clean and the two-model tests are exactly right. One thing worth folding in before this lands though, since the skip is macOS only and that's the platform a few other spots still describe the old way.

In the remote setup path, right after expose succeeds we still print a note saying the lerd-dns-forwarder is on lanIP:5300 answering star.test. On macOS that forwarder never gets installed now, so every Mac operator is told a service is listening that doesn't exist. The firewall line just below it is actually still fine because on macOS lerd-dns itself binds lanIP:5300, so the port really does need opening, it's only the forwarder attribution that's wrong. Probably cleanest to phrase that note off lerdDNSBindsLANPort so macOS says lerd-dns binds the LAN address directly.

Same staleness lives in the lan expose long help, it still says expose starts the userspace lerd-dns-forwarder and frames 5300 as the forwarder's port, so a macOS user reading help goes looking for something the design now omits.

Last one is lower stakes, DisableLANExposure still tears the forwarder down unconditionally even though enable now gates it. It's harmless today only because every teardown error is swallowed, but it leaves the enable and disable sides reasoning about the forwarder differently. A small ensureLANForwarderRemoved mirror gated on the same predicate would keep both halves honest.

None of this is a blocker on the core fix, just feels worth closing the loop on the same platform this targets so macOS folks aren't chasing a service that isn't there.

prabowosd:
Thanks, all three folded in (c07dcb4).

1. **Remote-setup note**: now phrased off `lerdDNSBindsLANPort`. macOS prints "lerd-dns binds <ip>:5300 directly, answering *.test", Linux keeps the forwarder wording. Left the firewall line as-is since the port is genuinely bound on both platforms.

2. **`lan expose` long help**: reworded to state the split explicitly. Linux bridges through the userspace lerd-dns-forwarder; macOS binds lerd-dns to the LAN address directly so no forwarder is installed.

3. **DisableLANExposure**: extracted to `ensureLANForwarderRemoved`, the named mirror of `ensureLANForwarder`, so both halves reference the same step.

One deliberate divergence on #3: I kept the teardown **unconditional** rather than gating it on `lerdDNSBindsLANPort`. A Mac that ran `lan expose` on a pre-this-PR build already has a real `lerd-dns-forwarder` installed (and on macOS it double-binds lanIP:5300 and breaks `.test` resolution). `unexpose` is the recovery path for those users, so it has to clear the stale unit even though the current macOS code never installs one. Gating removal would strand exactly the machines this PR is meant to fix. Removing a missing unit is ignored, so it stays a safe no-op on a clean macOS install. Documented that rationale on the function. Happy to gate it if you'd rather not carry the migration cleanup there.

`go build -tags nogui ./...`, `go vet`, and the cli tests are green.

PR #633, opened 2026-06-25
Port-ownership guard: coexist with a host-installed database server

iabduul7:
## Summary

This splits the **port-ownership guard** out of #588 into its own small PR, as we agreed in the #588 discussion. It lets lerd's containerized database coexist on one machine with a **host-installed** server of the same engine (system MySQL/MariaDB/Postgres) without a port fight. That's the one case the existing `.env.lerd_override` + `LERD_EXTERNAL_SERVICES` recipe can't cover.

When a DB engine's default published port (3306, 5432) is already owned by a host server, lerd's container can't bind it. The collision is worst at boot: systemd autostarts the quadlet with no lerd process around to step aside, so whichever unit starts first grabs the port and the other one fails. This guard runs at quadlet **write** time, so the port baked into the `.container` file (and therefore boot autostart) never collides in the first place.

## What it does

- **Auto-shift guard.** When lerd writes a DB preset's quadlet, if a host server owns the engine's default port and the user set no override, lerd publishes its container on the next free port instead, persists that choice, and rewrites the connection URL. The container-internal port doesn't change, so containerized apps (which reach the service by name over the lerd network) are unaffected.
- **Liveness-independent detection.** A host server counts as "in the way" if it's installed at all, meaning a populated data/cluster dir (`/var/lib/mysql`, `/etc/postgresql/<v>/…`) or a live socket, not just if it happens to be running at that moment. A cleanly stopped host DB still reclaims its port on the next boot, so the guard has to avoid it regardless of current run state. A plain "is the port listening" check would miss this.
- **`lerd service port <svc> [port] | --reset`.** A manual companion to move a service's published port (say, to free 3306 for a host MySQL) or reset it to the default. It pre-flights for a busy port, restarts the unit, and reports the port it actually ended up on.
- **Host-proxy follow.** Host-proxy sites connect over the published loopback port, so when that port moves their `.env` is regenerated to follow it. Container sites are left alone. Both the manual command and the auto guard trigger this.

## Scope

Kept deliberately small. Out of scope, staying with the shelved `db.external` work: the committed external-DB backend, host-mode env emission, the backend toggle, and all dashboard/UI. Per the split agreement the service page is unchanged, and the override recipe keeps its docs page.

## Relation to #628

#628 added an install-time warning when a host port is already busy, which is a nice heads-up at `lerd install`. This guard is complementary rather than overlapping: that check is advisory and runs at install, while this one actually moves lerd's published port at quadlet-write time, so the collision is avoided even at boot, where nothing is around to warn and systemd just starts the unit. The two don't touch the same code (the install warning lives in `install.go`, which this PR doesn't change).

## Review items folded in

This picks up the actionable items from @prabowosd's review on #588: reserve preset ports in the free-port search, fail closed if the override can't be persisted, a dual-stack (v4/v6) bindability check shared by the guard and the CLI pre-flight, glob `.s.PGSQL.*` so a Postgres cluster on a non-default port is still detected, and a warning when no free port is available.

## Testing

`gofmt`, `go build`, `go vet`, and `go test ./...` all green. There's unit coverage for port-owner attribution, dual-stack bind, reserved-port accounting, fail-closed persistence, install detection (socket file vs populated vs server-less marker dirs, and a non-default-port socket), the published-port override end to end through the host-proxy env rewrite, and the published-port-aware startup checks. I also smoke-tested it live: the CLI pre-flight correctly refused a port already held by a running host Postgres.

Supersedes the guard portion of #588, which can be closed once this lands.


prabowosd:
Well ill check asap thanks for PRs

geodro:
Thanks for this, the core machinery is solid, the bindability check, the free port search, the reserved port accounting and the host proxy follow are all the right pieces. I want to change the design a bit before it lands though, mostly to make it simpler and to stop it from looking at the host machine.

The thing I want to drop is that the guard inspects the host to decide, it stats socket paths, reads /var/lib/mysql and /etc/postgresql, and looks for the mysqld binary. lerd shouldn't be probing files or binaries on someone's computer at all, it's fragile and more than we should do, and on Debian and Ubuntu the rootless user usually can't even read /var/lib/mysql so it silently fails for the most common engine anyway. The only signal we should use is whether the port is free.

So make it generic and per service. When we write a service quadlet, check whether that service's own published port is bindable, and if it isn't, shift to the next free one and persist it, driven purely by port availability, never by what's installed on the host. That works for every service, redis, meilisearch, custom services, not just the three database families, and it lets us delete dbprobe.go and hostbackend.go and the whole detection layer.

Two rules on top of that. We only auto pick when a service has no port recorded yet, the published_port equals zero gate you already have, and once we pick one it sticks, we persist it and never move it again on our own, not even back to the default when that frees up later. If the user wants a different port after that they run lerd service port. No reshuffling on every start.

We also already have this allocator. Host proxy sites pick a free port for their Node and Vite dev servers in internal/cli/hostproxy.go with firstFreePort and portBoundOnHost, and it already does exactly this, it checks the port on v4 and v6 loopback and nothing else, and persists the choice so it stays stable. Could we reuse that instead of adding firstFreeHostPort and portBindable next to it, one shared free port helper for the dev servers and the services both.

The one consequence is that a host server that isn't running when we write the quadlet looks free, so we might take its port and clash on the next boot. That's the right trade for not snooping the host, a host database is usually running and the failure is loud, we just note it in the docs as a known limitation.

A few smaller things while you're in here. There's no docs page for any of this and we document every feature before it lands, so a short section on the auto shift, the service port command and reset. The control is also CLI only, it should be reachable through MCP too, a port action on the service tool. The .gitignore picks up a local data/ entry that looks like your working board, could you drop it or move it to your local exclude. And it's conflicting with main, so a rebase would help us review the merged result.

Really nice work, this is close, I just want it to be port availability only, generic, and built on the allocator we already have before we merge.

iabduul7:
Thanks, this all makes sense and I'm happy to take it in this direction. Port-availability only, generic, and built on the allocator we already have is clearly the better design, and it deletes more than it adds.

**Dropping host inspection.** Agreed completely. I'll remove `dbprobe.go` and `hostbackend.go` and the whole detection layer. You're right that statting `/var/lib/mysql` and `/etc/postgresql` and looking for `mysqld` is both overreach and unreliable, the rootless-user-can't-read-`/var/lib/mysql` point on Debian/Ubuntu is the clincher since that's the most common engine. The guard will key off one signal only: can we bind the port.

**Generic, per-service.** The guard moves to quadlet-write time for any service: if the service's own published port isn't bindable, shift to the next free one and persist. No DB-family special-casing, so redis, meilisearch and custom services get it for free, and the connection-URL rewrite stops being DB-specific.

**The two rules.** Auto-pick only when nothing is recorded yet (the `published_port == 0` gate stays), and once chosen it sticks, persisted and never moved automatically again, not even back to the default when it later frees. Changing it stays a deliberate `lerd service port`. No reshuffling on start.

**Reusing the allocator.** Will do. I'll lift `firstFreePort`/`portBoundOnHost` into one shared free-port helper used by both the dev servers and the services, and drop the `firstFreeHostPort`/`portBindable` I'd added alongside them.

**Boot-collision trade-off.** Accepted, a stopped host server looks free at write time so we might take its port and clash on next boot. I'll call it out as a known limitation in the docs rather than trying to detect around it.

**Docs.** I'll add a short page covering the auto-shift behaviour, the `lerd service port` command and `--reset`, plus that known limitation.

**MCP.** Good catch, I'll add a port action on the service tool so it's not CLI-only.

**Housekeeping.** Dropped the `data/` `.gitignore` commit (moved it to my local exclude), and rebased onto main, the only conflict was the `quadlet_embed.go` import after your podman single-seam change, resolved in favour of `execCommand`.

I'll push the reworked branch shortly. Thanks for the thorough review.

iabduul7:
Pushed the reworked branch. It's port-availability only and generic now, built on the shared allocator, and it drops the whole host-inspection layer.

**Dropped host inspection.** `dbprobe.go` and `config/hostbackend.go` are gone, along with `hostServerInstalled`/`hostOwnsDBPort`. Nothing stats sockets, reads `/var/lib/mysql` or `/etc/postgresql`, or looks for a binary anymore. The only signal is whether the port can be bound.

**Generic, per-service.** The guard moved to `EnsureCustomServiceQuadlet`, the choke point every quadlet passes through (the DB presets funnel in via `EnsureDefaultPresetQuadletPinned`). At write time, if a service's own primary host port can't be bound, it shifts to the next free port and persists. No DB-family special-casing, so redis, meilisearch and custom services get it too, and the connection-URL rewrite is no longer DB-specific.

**The two rules.** Auto-pick only when nothing is recorded (`published_port == 0`), and once picked it sticks: persisted, never moved again on its own, not even back to the default when it later frees. Changing it stays a deliberate `lerd service port`.

**Reused the allocator.** New leaf package `internal/freeport` with `Bindable` + `FirstFree`, used by both the host-proxy dev servers (`allocateHostPort`) and the service guard. The duplicate `firstFreePort`/`portBoundOnHost` (cli) and `firstFreeHostPort`/`portBindable` (serviceops) are gone. I took the race-free dual-stack bind (the v4 listener held open through the v6 bind) as the shared version since it's the stricter of the two.

**One thing worth a look.** The guard skips the shift while the service's own unit is active. `EnsureCustomServiceQuadlet` also runs during family regeneration, config edits and updates while a service is up, and a running service's own published port reads as busy to a bind probe, so without that check it would shift a healthy service off its own port on the next regeneration. It only reads our own unit's status, nothing on the host, so it stays port-availability only. A service that's down when its port has been taken still shifts correctly, which is the case we actually care about. The `lerd service port` command stops the unit before re-rendering for the same reason, so `--reset` onto a port a host server now owns re-shifts cleanly instead of failing to bind.

**Boot-collision limitation.** Documented as you suggested: a host server that's stopped at write time looks free, so we may take its port and clash on its next start. Recover with `lerd service port <name> <port>`.

**Housekeeping.** Docs section added (the auto-shift, `lerd service port`, `--reset`, and the limitation), MCP gets a `port` action on the service tool, the `data/` gitignore entry is dropped (moved to my local exclude), and the branch is rebased.

Issue #638, opened 2026-06-26
PHP images always build from source on podman < 5: `podman pull --policy` is rejected and the error is swallowed


geodro:
tryPullBaseImage in internal/podman/build.go pulls the prebuilt base image with `podman pull --policy=always`, and sets `cmd.Stderr = io.Discard`. The `--policy` flag on `podman pull` only exists from podman 5.0 onward. On any older podman the command fails with "unknown flag: --policy", and because the stderr is discarded the failure is invisible. lerd then silently falls back to the full local build even though the prebuilt base image is published, matches the version's Containerfile hash, and may already be present locally. Reproduced on podman 4.9.3 (the Ubuntu 24.04 LTS default): the exact published tag pulls fine by hand without `--policy`, but lerd builds the whole image from source. The `--policy=always` was introduced in 34fb74ae to force a fresh base pull on php:rebuild.

The blast radius is the entire Debian/Ubuntu stable ecosystem, not just old releases: Ubuntu 24.04 LTS ships podman 4.9.3, Ubuntu 22.04 ships 3.4.4, Debian 12 ships 4.3.1, all below 5.0. Only Arch, Fedora, and Ubuntu 25.04+ have podman 5.x. So the large majority of users silently rebuild every PHP image from source instead of pulling, which is slow, piles up disk (see #637), and on a fragile rootless storage backend tips into the `tar: Cannot change mode: Operation not permitted` build failure (see #635). The silent fallback also hides any other pull problem, a network reach to ghcr, a missing tag, a storage error, all surface as a confusing local build wall rather than a clear reason.

lerd already knows this version boundary: version.go gates the quadlet `StopTimeout=` key behind podman 5.0 with the note that Ubuntu 24.04 ships 4.9.3. The same gate is needed here.

Fix direction: stop discarding the pull stderr so a real failure is visible, and either gate `--policy` behind a podman 5.0 check or drop to a form every supported podman accepts. Pairs with the podman version preflight in #636. Refs #635.

PR #639, opened 2026-06-26
release: v1.26.1

geodro:
This is the v1.26.1 patch release. The headline fix is that lerd was silently rebuilding every PHP image from source on most Linux hosts: it pulled the prebuilt base with `podman pull --policy=always`, but `--policy` only exists on podman 5.0, so on Ubuntu 24.04 (4.9.3), 22.04 (3.4.4) and Debian 12 (4.3.1) the flag was rejected, and because the pull's stderr was discarded the failure was invisible and lerd fell back to the full local build even when a matching prebuilt base was published. The flag is now gated behind a podman 5.0 probe and the pull error surfaces in the fallback message.

Alongside that it cherry-picks five already-merged fixes from main worth shipping now rather than waiting for the next feature release: setup reports npm and bun build failures instead of killing the process, host workers whose deps aren't installed surface a remedy instead of dropping silently and update's restarts route through the feedback layer, macOS LAN exposure keeps its published ports and stops double-binding the DNS forwarder, and the host database override docs example is corrected.

Refs #638.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,18 @@ Lerd uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ---
 
+## [1.26.1] - 2026-06-26
+
+A patch release fixing a silent fallback that made most Linux hosts rebuild every PHP image from source, alongside cherry-picked fixes for macOS LAN exposure, setup feedback, and host-worker reporting.
+
+### Fixed
+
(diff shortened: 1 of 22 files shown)


geodro:
Superseded by #640. That one carries the same content plus the simpler drop-the-flag pull fix from 1.26.2, resolved against main's execCommand seam, so closing this in favour of it.

PR #641, opened 2026-06-26
fix(mcp): honor host-proxy sites in site link

### LINKED ISSUE #642: MCP `site link` downgrades host-proxy sites to PHP
## Summary

Calling `site link` through the MCP server on a **host-proxy** site silently downgrades it to a plain PHP-FPM site and breaks it (403).

## Cause

The MCP `site link` handler (`execSiteLink` in `internal/mcp/server.go`) only branches on two cases: custom-container (`proj.Container`) and PHP/framework. A `.lerd.yaml` carrying a `proxy:` section has neither, so it falls through to the **PHP path**. The CLI `lerd link` (`internal/cli/link.go`) handles host-proxy correctly; only the MCP handler is missing the branch.

## Impact

After an MCP `site link` on a host-proxy site:

- the registry entry (`sites.yaml`) loses `host_port` / `host_command` and gains `php_version`,
- the generated vhost loses `proxy_pass` and becomes a plain FPM vhost serving `public/`,
- the site returns 403, and `site restart` then reports it as "static".

## Reproduce

1. A host-proxy site (`.lerd.yaml` with a `proxy:` block, e.g. a Nuxt/Vite dev server on a host port), linked and serving.
2. Run `site link` via the MCP server (e.g. to re-apply config).
3. The proxy vhost is replaced by an FPM vhost and the site 403s.

Recovery today: re-link with the CLI, `lerd link` (which honors the `proxy:` block).

## Fix

PR #641 adds a host-proxy branch to `execSiteLink` that delegates to `lerd link` (same shell-out pattern as `execWorkerStart`), reusing the cli package's consent gating and dev-server supervision. No `--yes`, so the consent gate still applies.

prabowosd:
Fixes #642

## Problem

The MCP `site link` handler (`execSiteLink` in `internal/mcp/server.go`) only branches on two cases: custom-container (`proj.Container`) and PHP/framework. A `.lerd.yaml` carrying a `proxy:` section has neither, so it falls through to the **PHP path**.

The result is that calling `site link` (via MCP) on a host-proxy site silently downgrades it:

- the registry entry loses `host_port` / `host_command` and gains `php_version`,
- the generated vhost loses `proxy_pass` and becomes a plain FPM vhost serving `public/`,
- the site starts returning 403 (and `site restart` then reports it as "static").

The CLI `lerd link` (`internal/cli/link.go`) handles host-proxy correctly; only the MCP handler was missing the branch. Reproduced on a Nuxt dev site (`pnpm dev` on a host port): one MCP `site link` turned a working host-proxy site into a broken FPM site.

## Fix

Add a host-proxy branch to `execSiteLink` that delegates to `lerd link`, mirroring the shell-out pattern worker start (`execWorkerStart`) already uses. Delegating reuses the cli package's consent gating and dev-server supervision rather than reimplementing them in the MCP package (which can't import `cli`).

Consent is preserved, not bypassed: the subprocess `lerd link` is non-interactive, so the existing `hostProxyGate` lets an already-approved command (or `host_proxy.skip_confirmation`) proceed, while an unapproved command is refused with the usual guidance instead of run blindly. The MCP deliberately does not pass `--yes`.

## Notes

- No new test: the branch is a guard + shell-out (same shape as `execWorkerStart`, which is likewise not unit-tested for the shell-out), and the consent policy itself is already covered by the `hostProxyGate` unit tests. Mocking `runIn` here would test the mock, not behavior. Happy to add one if you'd prefer a specific shape.
- `go vet` / `go build` / `go test ./internal/mcp/` (59 pass) all green with `-tags nogui`.

### DIFF
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -2437,7 +2437,12 @@ func execSiteLink(args map[string]any) (any, *rpcError) {
 	// proceeds, while an unapproved command in this non-interactive context is
 	// refused with guidance rather than run blindly.
 	if proj != nil && proj.Proxy != nil && proj.Proxy.Port > 0 {
-		out, err := runIn(projectPath, "lerd", "link", name)
+		// No positional name: a positional is treated by runLink as an explicit
+		// primary domain to prepend, which would register the directory-derived
+		// name alongside the .lerd.yaml domains (and SyncProjectDomains would then
+		// persist the spurious entry). Plain `lerd link` honors proj.Domains
+		// verbatim, matching the container and PHP branches above.
+		out, err := runIn(projectPath, "lerd", "link")
 		if err != nil {
 			msg := strings.TrimSpace(out)
 			if msg == "" {
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -2427,6 +2427,27 @@ func execSiteLink(args map[string]any) (any, *rpcError) {
 		return toolOK(fmt.Sprintf("Linked %s -> %s (custom container, port %d)", name, strings.Join(domains, ", "), proj.Container.Port)), nil
 	}
 
+	// Host-proxy path: .lerd.yaml has a proxy section, so the site runs a dev
+	// server on the host that nginx reverse-proxies to. Supervising that command
+	// needs the consent gating and worker-start logic that live in the cli
+	// package, so delegate to `lerd link` (the same shell-out pattern as worker
+	// start) instead of falling through to the PHP path, which would downgrade
+	// the site to plain FPM and drop the proxy vhost. The consent gate still
+	// applies: an already-approved command (or host_proxy.skip_confirmation)
+	// proceeds, while an unapproved command in this non-interactive context is
+	// refused with guidance rather than run blindly.
+	if proj != nil && proj.Proxy != nil && proj.Proxy.Port > 0 {
+		out, err := runIn(projectPath, "lerd", "link", name)
+		if err != nil {
+			msg := strings.TrimSpace(out)
+			if msg == "" {
+				msg = err.Error()
+			}
+			return toolErr(msg), nil
+		}
+		return toolOK(strings.TrimSpace(out)), nil
+	}
+
 	// PHP / framework path.
 	framework := ""
 	if fname, ok := config.DetectFrameworkForDir(projectPath); ok {


geodro:
Thanks for this. I built the branch locally and ran it end to end, and the core fix is solid: linking a host-proxy project through the MCP site_link tool now registers it as a proxy site with host_port set and generates a proxy_pass vhost instead of the old FPM downgrade. With a real upstream bound to the port, a request through nginx returns 200 and the body is proxied through, so the path the PR set out to fix genuinely works.

There is one issue I'd want sorted before merge, and testing made it concrete. Because the branch shells out `lerd link <name>` with the directory-derived name as a positional argument, and runLink treats the first positional as an explicit domain to prepend as primary (link.go:203-213), the registered domains come out wrong. I linked a project whose .lerd.yaml says `domains: [shop641]` from a directory named nuxtshop641, and it registered as nuxtshop641.test plus shop641.test, with the directory-name domain promoted to primary. Running plain `lerd link` on the same project, with no positional argument, registers only shop641.test, which confirms the positional name is the cause. The container and PHP branches in this same function honor proj.Domains verbatim, so the proxy branch diverges from them. It's also a bit worse than it first looks: the handler then calls SyncProjectDomains, which rewrote the project's .lerd.yaml on disk to `domains: [nuxtshop641, shop641]`, so the spurious domain persists and later links inherit it. Dropping the positional name, or passing the resolved domains explicitly, should line this back up with the other two branches.

Worth flagging the consent path too. The delegated `lerd link` runs in a non-interactive subprocess with no TTY, so a host-proxy command that hasn't been approved yet hits the gate's refusal and the site is left entirely unlinked. As a gate that's reasonable, but for a brand-new site it dead-ends the user out to a terminal with no way to finish from inside the editor, so I wanted to check whether that's the intended flow.

On worktrees, behavior is fine but the messaging is inconsistent. Running it on a worktree path doesn't register a bogus separate site, but it surfaces as a domain-collision error ("domain X is already used by site Y") rather than the clear worktree explanation the CLI prints, because the worktree inherits the parent's domains and the handler's own domain pre-check fires before the new branch is reached. Not a correctness problem, just a confusing message compared to `lerd link` in a shell.

prabowosd:
Thanks for building it end to end, the positional-name bug is a real one and I've pushed a fix (2c2c4c62).

**Positional name (the blocker):** dropped it. The proxy branch now shells out plain `lerd link` with no positional, so runLink honors `proj.Domains` verbatim instead of prepending the directory-derived name as primary. That lines it up with the container and PHP branches, and removes the SyncProjectDomains write-back that was persisting the spurious domain. Re-tested on `nuxtshop641` with `domains: [shop641]`: now registers only `shop641.test`.

**Consent dead-end:** yes, that's the intended gate, but I take your point that dead-ending a brand-new site out to a terminal is a poor exit. The refusal text already carries the approve-and-rerun guidance, so the user isn't stranded blind, but I'd rather not bypass the gate from MCP. Happy to follow up with a clearer 'approve then re-link' hint in the refused-from-MCP case if you think it's worth it, but I'd keep it out of this PR.

**Worktree collision message:** confirmed not a correctness issue, and it's the shared domain pre-check (the same one all three branches sit behind) firing before the proxy branch, not anything this PR introduced. The CLI's nicer worktree explanation lives downstream of that check. Leaving it here to keep the diff to the actual fix, can split a separate message-clarity change if you want it.

Issue #659, opened 2026-06-27
idle-suspend orphans in-container workers (file-watcher.cjs) on macOS — suspended sites keep burning CPU


prabowosd:
## Summary

On macOS, when idle-suspend stops a site's workers, the in-container worker process (and its children) survive as orphans. The site is marked `idle_suspended: true` but the `node file-watcher.cjs` auto-reload watcher (spawned by `horizon:listen` / `octane:start --watch` when `horizon:reload`/`octane:reload` is on) keeps running and keeps polling, so an "idle" site still burns significant CPU.

## Evidence (live, v1.26.2)

Two sites with `horizon:reload on`, both reported idle+suspended by the engine:

```
$ curl -s :7073/api/sites   # both: idle=true, idle_suspended=true
lferp idle=True suspended=True | erprajarepair idle=True suspended=True

$ podman top lerd-php84-fpm pcpu args | grep file-watcher.cjs
23.730   node file-watcher.cjs [...erprajarepair...]
$ podman top lerd-php85-fpm pcpu args | grep file-watcher.cjs
25.340   node file-watcher.cjs [...lferp...]
```

Suspended for >4 minutes, watchers never died, ~25% CPU each (≈50% of a core continuously) on an otherwise idle laptop.

## Root cause

On macOS workers run in podman-exec mode: launchd runs `podman exec -w <sitePath> ... php artisan horizon:listen`, which spawns `node file-watcher.cjs` as an in-container child. `stopWorkerUnit` (`internal/cli/worker.go`) does `StopUnit` (launchctl unload) + `RemoveServiceUnit` + `removeWorkerExecArtifacts` (drops the .sh/.pid). Stopping the launchd job kills the **host-side** `podman exec`, but the **in-container** `horizon:listen` (and its `node` child) are never signaled, so they keep running.

The cwd-scoped in-container reap logic exists in `buildWorkerGuard` (`internal/cli/worker_guard.go`) but only runs on **launch** (to clear orphans before respawn), never on **stop**. So a stop-without-restart (exactly what idle-suspend does) leaves the orphan alive until the next start.

## Proposed fix

[...]

PR #661, opened 2026-06-27
fix(cli): reach lerd-ui over TCP on macOS so idle/dump status works

### LINKED ISSUE #660: lerd idle status always shows '(lerd-ui not running)' on macOS
## Summary

`lerd idle status` always reports `(lerd-ui not running)` for every site on macOS, even when lerd-ui is running and idle-suspend is fully functional.

## Root cause

`fetchIdleSites()` (`internal/cli/idle.go`) reads `/api/sites` over the lerd-ui **unix socket** via `getUnix` → `config.UISocketPath()` (`lerd-ui.sock`). But that socket is created **Linux-only**: `internal/ui/server.go` guards the unix listener with `if runtime.GOOS != "darwin"` (the socket is for the lerd.localhost nginx vhost, which doesn't apply on macOS). So on macOS `lerd-ui.sock` never exists, `getUnix` fails, and `idleSiteStatus` returns `(lerd-ui not running)`.

The daemon IS reachable on macOS over HTTP at `127.0.0.1:7073` (the dashboard uses it, and `:7073/api/sites` returns the full idle state including `idle`, `idle_suspended`, `last_active`). Only the CLI status path uses the unavailable unix socket.

## Proposed fix

Make the CLI's daemon read fall back to the HTTP loopback (`127.0.0.1:<ui port>`) when the unix socket is unavailable (i.e. on macOS), so `lerd idle status` works the same as the dashboard. `idle on/off/timeout` already work on macOS (they go through the idle-control socket), so this is scoped to the status read.

PR incoming.

prabowosd:
## Problem

`lerd idle status` reports `(lerd-ui not running)` for every site on macOS even when lerd-ui is up and idle-suspend is working.

## Root cause

The CLI's daemon client (`getUnix`/`postUnix`/`dialUnixHTTP` in `internal/cli/dump.go`) always dials the lerd-ui **unix socket** (`config.UISocketPath()`). But that socket is created **Linux-only**: `internal/ui/server.go` guards the unix listener with `if runtime.GOOS != "darwin"` (it exists for the lerd.localhost nginx vhost). So on macOS `lerd-ui.sock` never exists, the dial fails, and `idleSiteStatus` returns `(lerd-ui not running)`.

The daemon is reachable on macOS over HTTP at `127.0.0.1:7073` (the dashboard uses it; `:7073/api/sites` returns full idle state). Only the CLI status path used the unavailable socket. `lerd dump`'s SSE tail had the same problem.

## Fix

Add `config.UIClientNetwork()` / `config.UIClientAddr()` (mirroring the existing `DumpsListenNetwork`/`DumpsListenAddr` darwin split): TCP `127.0.0.1:7073` on macOS, the unix socket on Linux. The CLI client now uses them, so it reaches the daemon the same way the dashboard does. Linux behavior is unchanged.

Scope: read/status only. `idle on/off/timeout` already work on macOS (they go through the idle-control socket, which is created on both OSes).

## Test

`TestUIClientTransport_matchesOS` asserts the transport matches the host OS. `go build ./...` clean.

Note: this is the cosmetic half of the macOS idle-suspend issues. The functional one (suspended sites still burning CPU because the in-container worker/watcher is orphaned) is #659.

Fixes #660.


geodro:
Confirmed the root cause, the unix socket really is bound Linux only while the daemon always listens on tcp 7073, so dialing the loopback on macOS is the right call and matching the DumpsListen split keeps it consistent. Linux path is untouched and the var injection keeps the socket test honest on both platforms. Reads clean to me.

One thing I'd flag for a follow up, the same socket dial still lives in the TUI dump tail and the MCP dump tail, so a macOS user hitting dumps through either of those gets the identical failure this fixes for the CLI. Out of scope for #660 since that's specifically idle status, but worth a separate issue so it doesn't get lost.

Tiny nit, the new uiClientDial var landed between the dialUnixHTTP doc comment and the function, so that comment now sits above the var and still says it opens a conn over the unix socket which isn't true on macOS anymore. Worth nudging the var below the comment or rewording it.

prabowosd:
Fixed the nit in `01a3413`: moved `uiClientDial` above its own doc comment and reworded `dialUnixHTTP`'s comment so it no longer claims the dial always goes over the Unix socket (it now describes whichever transport `uiClientDial` resolves, unix on Linux, TCP loopback on macOS).

On the follow-up: agreed the TUI dump tail and MCP dump tail still dial the Unix socket and hit the identical failure on macOS. Out of scope for #660 (idle status), so I'll open a separate issue to track fixing both dump-tail paths to use `uiClientDial` the same way.

PR #662, opened 2026-06-27
fix(worker): reap in-container worker on stop (macOS) so idle-suspend frees CPU

### LINKED ISSUE #659: idle-suspend orphans in-container workers (file-watcher.cjs) on macOS — suspended sites keep burning CPU

prabowosd:
## Problem

On macOS, idle-suspend marks a site `idle_suspended: true` and stops its worker units, but the in-container worker process and its children survive. The `node file-watcher.cjs` auto-reload watcher (spawned by `horizon:listen` / `octane:start --watch` when reload is on) keeps polling, so an "idle" site still burns ~25% CPU per site.

Observed live (v1.26.2), two sites reported idle+suspended by the engine, watchers never died over 4+ minutes:

```
curl -s :7073/api/sites  → siteA idle=True suspended=True | siteB idle=True suspended=True
podman top lerd-php85-fpm → node file-watcher.cjs [...siteA...]  25.3% CPU
podman top lerd-php84-fpm → node file-watcher.cjs [...siteB...]  23.7% CPU
```

## Root cause

macOS workers run in podman-exec mode: launchd runs `podman exec -w <sitePath> <fpm> php artisan <worker>`. Stopping the launchd job (`stopWorkerUnit` → `StopUnit` → launchctl unload) kills only the **host-side** `podman exec`. The process it started **inside** the FPM container, and its `file-watcher.cjs` child, are never signaled, so they keep running.

The cwd-scoped in-container reap already exists in `buildWorkerGuard` (`internal/cli/worker_guard.go`) but only runs on **(re)launch**, to clear orphans before a respawn. A stop-without-relaunch, exactly what idle-suspend does, never reaps, so the orphan lingers until the next start.

## Fix

- Factor the launch guard's reap into `inContainerReapSnippet` (cwd-scoped `pgrep -f <workerCmd>` → SIGTERM, grace-wait, SIGKILL) and add `buildWorkerReapCommand` (`podman exec <container> sh -c <snippet>`).
- Persist that command as a per-worker `.reap` sidecar when the exec-mode worker is written (`writeWorkerExecUnit`).
- Run it on stop in `removeWorkerExecArtifacts` (macOS-only, already the stop hook), before dropping the `.sh`/`.pid`/`.reap` artifacts.

Cwd-scoping by `sitePath` keeps it from killing the same worker type in other sites that share the FPM container (the existing guard relies on the same disambiguation). Best-effort and time-boxed (10s) so a hung podman never blocks a stop. Workers started by an older build have no sidecar and are simply skipped. Linux is unaffected (`removeWorkerExecArtifacts` is a no-op there).

## Test

`TestBuildWorkerReapCommand_TargetsContainerAndCwd` executes the generated reap through a podman stub and asserts it calls `podman exec <container> sh -c` with the cwd-scoped kill for the right worker+site. Existing `buildWorkerGuard` tests still pass (the refactor is output-identical). `go vet` clean.

Pairs with #661 (the cosmetic half: `lerd idle status` on macOS). Fixes #659.


geodro:
Traced the integration and the placement is right. Baking the resolved container into the .reap sidecar at write time means stop just reads the finished command with no re-resolution, the snippet reuse across launch and stop is a clean refactor, and the cwd scoping keeps it from touching sibling sites in the shared FPM container. The stop ordering is the part I most wanted to check and it holds, Disable plus StopUnit unload the launchd job and kill the host side podman exec before the reap runs, so there's no window where the guard respawns the worker mid reap. Best effort, time boxed, older workers without a sidecar skipped, Linux a real no op. Reads well.

One thing I want to confirm before merging since the whole point is killing the file-watcher child. The snippet pgreps for the artisan command and filters by cwd, but the watcher's command line is node file-watcher.cjs which won't match that pattern, so the watcher only dies if killing the parent cascades to it. That cascade is probably real, the same snippet has run on relaunch in production without anyone reporting watcher buildup, but the new test only asserts the command shape, it doesn't prove the watcher exits, and the description shows the before state without an after. Could you confirm live that once a site is idle suspended on macOS, podman top on the fpm container no longer lists that site's file-watcher.cjs and the CPU actually drops? If you've already seen that, happy to ship.

prabowosd:
Confirmed live on macOS (v1.26.2 + this branch). A secured site with `reload_workers: [horizon]` on `lerd-php85-fpm`:

**Before stop** (`podman top <fpm> pid pcpu args`):

```
22   0.000    php artisan horizon:listen --poll
25   41.497   /usr/bin/node file-watcher.cjs ["/Users/.../<site>", ...]
36   0.000    php artisan horizon
```

**After the worker stop** (same `removeWorkerExecArtifacts` path idle-suspend funnels through): zero `file-watcher.cjs`, zero `horizon:listen`, zero `horizon` for that container. CPU back to idle.

On the cascade question: the watcher's cwd is `.../vendor/laravel/horizon/bin`, not the site path, so the cwd-scoped reap can't (and doesn't) target it directly. It dies via the SIGTERM-to-parent cascade, `ListenCommand`'s trap calls `$this->watcherProcess->stop()` when the `horizon:listen` parent is signalled, exactly as designed. The `.reap` sidecar pgreps `php artisan horizon:listen --poll` cwd-scoped to the site path, which matches only that parent. Good to ship.

One non-blocking note: the reap's grace is ~5s before SIGKILL. If the inner `horizon` ever takes >5s to drain, the parent could be SIGKILLed before its trap reaches `watcherProcess->stop()`, re-orphaning the watcher. For an idle site (no jobs draining) the trap returns effectively instantly, so it doesn't bite in the idle-suspend path this targets, just flagging it for the record.

PR #675, opened 2026-06-29
fix(watcher): bound source-file watches so an asset dump can't exhaust fds

### LINKED ISSUE #674: macOS: idle source-file watcher exhausts file descriptors on large source trees (EMFILE)
### Summary

The idle source-file watcher can exhaust the process file-descriptor limit, after which every file operation in lerd-watcher fails `too many open files` and unrelated watcher work breaks (e.g. idle-resume can no longer rewrite a vhost or persist the site registry).

### Root cause

`WatchSourceFiles` walks each site's source roots (`app`, `resources`, `routes`, ...) and `fsnotify.Add`s every directory. On macOS fsnotify's kqueue backend opens **one fd per file** in each watched directory. A generated or vendored asset dump checked into a source root, for example a full Font Awesome icon set (~150k SVGs under `resources/`), makes the watcher try to hold hundreds of thousands of fds.

Observed live: lerd-watcher holding ~92k fds, hitting `kern.maxfilesperproc` (92160), with the bulk being regular files under one site's `resources/` (a ~187k-file tree). Once at the cap, opens fail EMFILE and surface as e.g.:

```
[WARN] idle-resume host-proxy vhost <site>: generating host-proxy SSL vhost: open .../<domain>-ssl.conf: too many open files
[WARN] idle-resume persist <site>: open .../sites.yaml: too many open files
```

### Expected

The watcher should bound its fd footprint regardless of project structure, so a checked-in asset dump can't take down the whole watcher.

prabowosd:
Fixes #674.

## Problem

The idle source-file watcher walks each site's source roots (`app`, `resources`, `routes`, ...) and `fsnotify.Add`s every directory. On macOS fsnotify's kqueue backend opens **one fd per file** in each watched directory, so a generated or vendored asset dump checked into a source root (e.g. a Font Awesome icon set with ~150k SVGs under `resources/`) makes the watcher try to hold hundreds of thousands of fds. It hits `kern.maxfilesperproc` (~92k), and every open after that fails `too many open files`, which then breaks unrelated watcher work, e.g. idle-resume can no longer rewrite a vhost or persist `sites.yaml`:

```
[WARN] idle-resume host-proxy vhost <site>: ... <domain>-ssl.conf: too many open files
[WARN] idle-resume persist <site>: ... sites.yaml: too many open files
```

Live: lerd-watcher held ~92k fds (= `kern.maxfilesperproc`), almost all regular files under one site's `resources/` (a ~187k-file tree dominated by a checked-in icon set).

## Fix

Bound the watch set in `WatchSourceFiles` two ways:

- **Per-directory cap**: skip a directory (and its subtree) with more than `maxSourceDirEntries` (2000) direct entries. That is an asset dump, not hand-edited source, so kqueue's per-file fds don't pile up and activity detection loses nothing. Real source dirs hold tens to a few hundred files, far below the cap, so they stay watched.
- **Global backstop**: cap the total files registered across all targets at `maxWatchedSourceFiles` (32768), well under `kern.maxfilesperproc`, so no combination of trees can drive the watcher into EMFILE. The nginx access feed still drives activity for anything left unwatched (it is the primary signal anyway).

Skips are logged once per directory so the 30s rescans do not spam.

This complements the existing name-based excludes (`node_modules`, `vendor`, hidden dirs); those miss an asset dump sitting under a legitimate source root with an arbitrary name.

## Test

`TestWatchSourceFiles_skipsOversizedDir`: a dir with more than `maxSourceDirEntries` files under a source root reports no activity (skipped), while a normal source-file save in the same target still does (the skip is not over-broad). Existing `TestWatchSourceFiles_savesFireActivity` still passes. `go build`, `go vet`, and the full `internal/watcher` suite (64 tests) pass.

### DIFF
--- a/internal/watcher/sourcefiles.go
+++ b/internal/watcher/sourcefiles.go
@@ -19,6 +19,23 @@ type SourceTarget struct {
 	Dirs []string
 }
 
+// maxSourceDirEntries caps how many direct entries a directory may have before
+// the source watcher skips it. fsnotify's macOS (kqueue) backend opens a file
+// descriptor per file in every watched directory, so a generated or vendored
+// asset dump checked into a source root (e.g. a Font Awesome icon set with
+// thousands of SVGs) can exhaust the per-process fd limit. Such a directory is
+// not hand-edited source, so skipping it costs nothing for activity detection
+// while keeping fd use bounded. Real source dirs hold tens to a few hundred
+// files; this is far above that and far below an asset dump.
+const maxSourceDirEntries = 2000
+
+// maxWatchedSourceFiles is a hard ceiling on the files the source watcher will
+// register across every target, a backstop under the per-directory cap so no
+// combination of trees can drive the watcher into EMFILE (kern.maxfilesperproc
+// is ~92k on macOS). Once reached, further trees are left to the nginx access
+// feed for activity, which is the primary signal anyway.
+const maxWatchedSourceFiles = 32768
+
 // WatchSourceFiles watches each target's source directories recursively and
 // calls onActivity(key), debounced per key, when a source file under them is
 // written, created, or renamed — i.e. when you save while coding. Heavy or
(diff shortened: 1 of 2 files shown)


geodro:
This is a clean fix and the test is solid this time, it actually exercises the skip path and checks the cap isn't over-broad by confirming a normal save in the same target still fires. Build, vet, gofmt and the full watcher suite all pass here, and placing the new checks after the already-watched short-circuit is the right call so watched dirs don't get re-read on rescan.

One thing worth tightening. The oversized dir never lands in dirKey, so it never hits that short-circuit, and skipped only gates the log line, which means every 30s rescan does a full ReadDir of the exact 150k entry tree this fix exists to avoid touching, just to rediscover it's oversized and SkipDir again. A quick check for skipped before the ReadDir, returning SkipDir straight away, would keep the rescan off that tree entirely. Not a blocker since a periodic ReadDir is nothing next to holding the fds, just a cheap win.

prabowosd:
Good catch, tightened in 996601f. An oversized dir never lands in `dirKey`, so the already-watched short-circuit never covered it. Now `skipped` is read alongside `dirKey` up front and returns `SkipDir` straight away when set, so a rescan never re-`ReadDir`s that (150k-entry) tree, it just steps over it. The log stays once-per-dir (gated on the first-time transition). Build, vet, and the watcher suite still pass.

Issue #666, opened 2026-06-28
Pest browser doctor reports green when the installed Playwright is older than pest-plugin-browser requires


geodro:
lerd pest:browser doctor only checks that ./node_modules/.bin/playwright exists, it never checks its version. pest-plugin-browser pins a required Playwright version (currently 1.59.1 in PlaywrightNpmServer) and rejects anything older, so a user with an older playwright sees an all green doctor but lerd pest then throws PlaywrightOutdatedException, or PlaywrightNotInstalledException when the playwright cli cannot run at all and the version recheck fails outright. That is what #664 is hitting.

I reproduced both: a fresh setup with playwright 1.61.1 runs a browser test to green, and downgrading to 1.55.0 keeps the doctor green while lerd pest fails. The feature itself works, the doctor just gives false confidence.

Fix the doctor, and ideally the install flow, to read the installed playwright version and compare it against the floor pest requires, and to actually run ./node_modules/.bin/playwright run-server --version and surface its real output instead of a bare green check. The required version can be read from the vendored pest-plugin-browser PlaywrightNpmServer constant so it stays in sync. Pinning playwright at install time is also worth considering so a fresh setup cannot drift below the floor.

Refs #664

geodro:
Superseding this with #677. 664 turned out to be a boot failure on a perfectly up to date Playwright, so the real fix is wider than the version check, the doctor needs to actually boot the server and surface its real error. Folding the version-floor idea into that one and closing this in favor of it.

PR #682, opened 2026-06-29
fix(pest:browser): boot the Playwright server in doctor so a boot failure stops false-greening

### LINKED ISSUE #677: pest:browser doctor false-greens whenever the Playwright server fails to boot, not only when it is outdated
Follow up to #664. While chasing #664 it became clear the pest:browser doctor gap is wider than an outdated Playwright version. The reporter has playwright 1.61.1, well above the 1.59.1 floor pest-plugin-browser pins, the install is byte for byte identical to a healthy machine, the browser is shimmed correctly, and the doctor is all green, yet lerd test still throws PlaywrightNotInstalledException.

The reason is that pest-plugin-browser boots the server with ./node_modules/.bin/playwright run-server --host H --port P --mode launchServer and waits for it to print Listening on. If that process dies first, the plugin throws PlaywrightNotInstalledException and discards the real stderr. The same exception is reused whether playwright is genuinely missing or the server just failed to boot, so the user is told to npm install playwright even though it is already there. A version check, or even running run-server --version, does not catch this, because --version can succeed while the actual launchServer boot fails for an unrelated reason.

The fix is to make the doctor do what pest does: boot run-server in launchServer mode, wait for Listening on, and on failure surface the process's real stderr instead of a bare green check or a misleading exception. Reading the installed playwright version and comparing it against the floor from the vendored PlaywrightNpmServer constant should fold in as one of the checks, and pinning playwright at install time is still worth considering, but the headline fix is booting the server and showing the truth.

Refs #664

prabowosd:
Fixes #677.

## Problem

The `pest:browser doctor` checks the package, chromium, the `playwright` npm binary and the shim, then reports green. But pest-plugin-browser actually runs the suite by booting `./node_modules/.bin/playwright run-server --mode launchServer` and waiting for `Listening on`. If that boot dies for a reason none of the presence checks see, the plugin throws `PlaywrightNotInstalledException` and discards the real stderr, so the user is told to `npm install playwright` even though it is already there and byte-for-byte identical to a healthy machine. A `--version` check does not catch this, because `--version` can succeed while the `launchServer` boot fails.

## Fix

Add a final doctor check that boots `run-server` in `launchServer` mode the same way pest does, waits for the ready marker, and on failure surfaces the process's real output instead of a misleading green:

```
  [✓] playwright npm package installed
  [✓] Playwright browser shimmed to musl chromium
  [x] Playwright server boots (run-server, launchServer mode)
        → the server failed to start; its output is below
        | <the real run-server stderr>
```

The boot uses an ephemeral port (`--port 0`) so the probe never collides with a real run, and the process is killed the moment it is listening. The orchestration (`watchPlaywrightBoot`) is kept exec-agnostic so it is unit-tested with a plain `sh` command for the ready, failed (output surfaced), and timed-out cases.

## Scope

Doctor-only, no behaviour change to install/remove/test. The version-vs-floor fold-in mentioned in the issue (#666 territory) is left out to keep this focused on the headline boot check.

## Tests

`go test ./internal/cli/` (998 pass, including the new boot cases; race-clean).


prabowosd:
Validated end-to-end on a real PHP 8.5 FPM container (not just the unit tests).

**run-server invocation** — confirmed `playwright run-server --host 127.0.0.1 --port 0 --mode launchServer` prints `Listening on ws://127.0.0.1:<ephemeral>/`, so the marker and `--port 0` are right.

**Doctor, success path** — playwright present and bootable:
```
[✓] playwright npm package installed
[✓] Playwright server boots (run-server, launchServer mode)
```

**Doctor, the #677 case** — playwright present but the server boot fails (broke run-server while keeping the binary in place):
```
[✓] playwright npm package installed
[x] Playwright server boots (run-server, launchServer mode)
      → the server failed to start; its output is below
      | Error: browserType.launchServer: Executable doesn't exist at .../headless_shell
      |   Looks like Playwright Test or Playwright was just installed or updated.
```
The presence check stays green (the old trap) while the new check catches the boot failure and surfaces the real stderr instead of telling the user to reinstall playwright.

**One fix the live run surfaced:** killing the podman exec client did not reap the run-server it spawned inside the container, so the success path left a stray process. Added a `pkill` on the probe's unique `--port 0` command line after probing; re-verified no process lingers. Pushed as a follow-up commit.

PR #716, opened 2026-07-02
fix(mcp): bound podman hot-path calls and self-heal a stalled machine

### LINKED ISSUE #715: MCP tool calls hang after macOS sleep; add podman timeouts and podman-machine self-heal
On macOS, when the machine sleeps and wakes, the podman machine VM and its gvproxy networking are left suspended or stalled. The lerd MCP server is a single threaded stdio process that reads one JSON-RPC request, runs it to completion, then reads the next, with no per-call timeout and no cancellation. Every core MCP handler shells out to podman exec or podman inspect through the plain exec.Command seam in internal/podman/exec.go with no context deadline. So after a wake, the first podman call blocks forever, the serve loop never advances, and the whole MCP server goes unresponsive to Claude. A user hit exactly this, coming back to their Mac and finding lerd frozen when driving it through the agent. Worth noting the activity path is fine here, MCP calls do record activity and do wake idle-suspended sites, and the battery optimisations are Linux-only and inert on macOS, so neither of those is involved. The cause is purely the untimed podman shell-out on the request path.

The first piece is to bound the podman calls on the MCP hot path. We already have the podman.CmdContext plus context.WithTimeout pattern in use elsewhere (internal/nginx/manager.go, serviceops/migrate.go, log tailing), the MCP exec handlers just never adopted it. Wrapping execArtisan, execComposer, the ensureFPMStartedMCP to StartFPM ContainerRunning check at the top of internal/php/fpm_start.go, and the cache.go not-started inspect fallback converts a permanent hang into a fast surfaceable error instead of a freeze.

The second piece is self healing, so the timeout becomes a recovery rather than just a cleaner failure. Most of the machinery already exists on macOS. restartPodmanMachineForHeal does podman machine stop followed by ensurePodmanMachineRunning, which starts the VM, retries once, and polls podman ps -q until the API socket is live, which is the known post-sleep fix. gvproxy cannot be restarted on its own since it is a child of the VM, but lerd already repairs the broken port forwards by force removing and recreating the lerd containers via removeLerdContainersForGvproxyHeal. The heal-on-error shape is already established too, healOverlayCorruptionIfNeeded and healGhostContainersIfNeeded match an error signature, restart the VM, clear stale containers, retry once, then escalate with guidance. The reactive path is to treat a deadline-exceeded from an MCP podman call the same way, restart the machine, retry the call once, and only surface an error if it still fails, carrying over the existing cooldown and retry-once guards so a genuinely dead machine does not get stop-started in a loop.

There is also an optional proactive path for later. The DNS watcher already detects wake correctly on macOS through the monotonic-vs-wallclock resumed() check, but the heal it triggers is Linux-only nginx, so on macOS the wake is detected every tick and nothing acts on it. We could attach a macOS-only machine health probe to that existing resumed block and recover the VM on wake before the user even fires a call. The tradeoff is roughly 30s detection latency and spinning the VM on every laptop open even when lerd is not touched, so I would gate it carefully and treat it as a follow up to the reactive work rather than part of the first pass.

prabowosd:
## Problem

On macOS, when the host sleeps and wakes, the podman machine VM and its gvproxy networking are left suspended. The MCP server is a single-threaded stdio loop that runs each request to completion with no per-call timeout, and every core handler shells out to podman through the untimed `exec.Command` seam. So after a wake the first `podman exec`/`inspect` on the request path blocks forever, the serve loop never advances, and lerd goes unresponsive to the agent driving it.

The activity/idle-suspend path and the Linux-only battery optimisations are not involved here; the cause is purely the untimed podman shell-out on the request path.

## Fix

**Piece 1: bound the hot path.**
- New `podman.EnsureMachineResponsive` runs a bounded `podman ps -q` probe (10s) before an MCP handler shells out for real, and gates `StartFPM` (the first podman call on both `exec:artisan` and `exec:composer`). A healthy machine answers in well under a second, so this only fires on a real stall. Once it passes the VM is responsive, so the artisan/composer exec that follows won't hang, and legitimately slow commands (composer install, migrations) keep no artificial timeout.
- The `ContainerCache` non-started fallback (`defaultPollFn` and the `Running` check) is now bounded too, so status reads surface a fast empty snapshot instead of hanging.

**Piece 2: self-heal the stall.**
- On a failed probe, `EnsureMachineResponsive` restarts the machine once and re-probes, so a post-sleep freeze becomes a self-healed retry, or a fast surfaced error if the VM is truly dead. It carries a cooldown so a burst of calls against a genuinely dead machine doesn't stop-start it in a loop.
- The darwin heal is wired to the existing `restartPodmanMachineForHeal` (stop + `ensurePodmanMachineRunning`, the same recovery the service start pass uses). `MachineHeal` is nil on Linux, where there is no machine VM. Its feedback lands on stderr under the MCP server's existing `guardStdout`, so it never corrupts the JSON-RPC stream.

**Piece 3: proactive heal on resume.**
- The DNS watcher's resume tick (the existing wall-clock-vs-monotonic gap detection) also un-stalls the machine on macOS: `healMachineOnResume` → `defaultHealMachine` → the same `podman.EnsureMachineResponsive`, so the MCP/exec path is recovered before the next agent call rather than on it. Reusing the reactive probe+heal means both share one cooldown, and it only restarts the VM when the bounded probe actually fails, so it doesn't spin the VM on every laptop open. It's guarded by `isStopped`, so a deliberate `lerd stop` is never resurrected, and it's macOS-only (nil elsewhere).

## Tests

- `internal/podman/machineheal_test.go` covers healthy (no heal), stall→heal→recover, dead-machine retry-once error, cooldown blocks the second heal, and the nil-hook (Linux) path.
- `internal/watcher/dns_test.go` covers healing the machine only on the tick that detects a resume, never on the first or a normal-cadence tick.

All `internal/podman` + `internal/php` + `internal/watcher` tests pass.

Closes #715

### DIFF
--- a/internal/cli/machineheal_wire_darwin.go
+++ b/internal/cli/machineheal_wire_darwin.go
@@ -1,0 +1,10 @@
+//go:build darwin
+
+package cli
+
+import "github.com/geodro/lerd/internal/podman"
+
+// Wire the MCP/exec hot-path self-heal to the same machine restart the service
+// start pass uses. On a post-sleep stall, podman.EnsureMachineResponsive calls
+// this to stop+restart the VM and wait for its API socket, then retries once.
+func init() { podman.MachineHeal = restartPodmanMachineForHeal }
--- a/internal/php/fpm_start.go
+++ b/internal/php/fpm_start.go
@@ -29,6 +29,12 @@ func FPMInstalled(version, container string) bool {
 // callers add any user-facing progress. Shared by the CLI php/artisan/shell
 // commands and the MCP exec handlers so both auto-start the same way.
 func StartFPM(version, container string) error {
+	// Gate on a bounded machine probe so a post-sleep stall surfaces (and self-
+	// heals) fast instead of hanging the caller on the untimed inspect below.
+	// Once this passes the VM is responsive, so the exec that follows won't hang.
+	if err := podman.EnsureMachineResponsive(); err != nil {
+		return err
+	}
 	if running, _ := podman.ContainerRunning(container); running {
 		return nil
 	}
--- a/internal/podman/cache.go
+++ b/internal/podman/cache.go
@@ -49,10 +49,20 @@ func (c *ContainerCache) SetOnChange(fn func()) {
 	c.onChangeMu.Unlock()
 }
 
+// pollTimeout bounds the non-started fallback podman calls so a stalled VM
+// (macOS post-sleep) yields a fast empty snapshot instead of hanging the caller.
+const pollTimeout = 10 * time.Second
+
 func defaultPollFn() (string, error) {
-	return Run("ps", "-a",
+	ctx, cancel := context.WithTimeout(context.Background(), pollTimeout)
+	defer cancel()
+	out, err := CmdContext(ctx, "ps", "-a",
 		"--filter", "name=lerd-",
-		"--format", "{{.Names}}\t{{.State}}")
+		"--format", "{{.Names}}\t{{.State}}").Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
 }
 
 // Cache is the process-wide container state store. Start it once from serve-ui;
@@ -84,10 +94,21 @@ func (c *ContainerCache) Running(name string) bool {
 	c.mu.RUnlock()
 
 	if !started {
-		running, _ := ContainerRunning(name)
-		return running
+		return containerRunningBounded(name)
 	}
 	return v
+}
+
+// containerRunningBounded is the non-started fallback's container check, bounded
+// by pollTimeout so a stalled VM returns fast (false) instead of hanging.
+func containerRunningBounded(name string) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), pollTimeout)
+	defer cancel()
+	out, err := CmdContext(ctx, "inspect", "--format={{.State.Running}}", name).Output()
+	if err != nil {
+		return false
+	}
+	return strings.TrimSpace(string(out)) == "true"
 }
 
 // Snapshot returns a copy of the cached container map. When the cache has not
--- a/internal/podman/machineheal.go
+++ b/internal/podman/machineheal.go
@@ -29,10 +29,13 @@ var (
 // stall it heals once (cooldown-guarded) and re-probes, turning a post-sleep
 // freeze into a self-healed retry or a fast error instead of an unbounded hang.
 func EnsureMachineResponsive() error {
+	if MachineHeal == nil {
+		return nil // no machine VM to stall or heal (Linux, tests): skip the probe
+	}
 	if machineResponds() {
 		return nil
 	}
-	if MachineHeal == nil || !healOnceWithinCooldown() {
+	if !healOnceWithinCooldown() {
 		return fmt.Errorf("podman machine is not responding (try: lerd start)")
 	}
 	if machineResponds() {
@@ -45,9 +48,6 @@ func EnsureMachineResponsive() error {
 // returning whether a heal actually ran. Guards against stop-start loops when
 // the machine stays dead across successive calls.
 func healOnceWithinCooldown() bool {
-	if MachineHeal == nil {
-		return false
-	}
 	healMu.Lock()
 	if !lastHealAt.IsZero() && time.Since(lastHealAt) < healCooldown {
 		healMu.Unlock()
--- a/internal/podman/machineheal.go
+++ b/internal/podman/machineheal.go
@@ -1,0 +1,73 @@
+package podman
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+)
+
+// MachineHeal recovers a stalled Podman Machine (e.g. macOS post-sleep, when the
+// VM and its gvproxy networking are suspended and every podman call blocks
+// forever). Wired by the darwin CLI to restartPodmanMachineForHeal; nil on
+// Linux and in tests, where there is no machine VM to restart.
+var MachineHeal func()
+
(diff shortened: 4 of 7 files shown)


prabowosd:
Verified end-to-end on macOS against a real podman machine, both the happy path and the self-heal path.

**Happy path (machine up):** drove the MCP server directly with a `tools/call` for `exec:artisan --version`. The new `EnsureMachineResponsive` gate is a fast no-op (probe answers in well under a second), artisan runs through it, and there's no regression:

```
request:  exec:artisan --version   (path: a Laravel project)
response: {"result":{"content":[{"text":"Laravel Framework 13.12.0"}]}}
stderr:   (empty)
```

**Self-heal path (machine stopped to simulate a post-sleep stall):**

```
podman machine stop            → `podman ps -q` fails in 0.015s
MCP exec:artisan --version fired:
  EnsureMachineResponsive probe fails
  → MachineHeal → restartPodmanMachineForHeal:
      "restarting the Podman Machine to remount it…"
      "Starting Podman Machine…"
      "Waiting for Podman Machine to be ready… ready"
  → re-probe OK → StartFPM → artisan runs
response: {"result":{"content":[{"text":"Laravel Framework 13.12.0"}]}}
total:    15.7s (recovered, not an unbounded hang)
```

Confirmed:
- The stall recovers automatically instead of freezing the serve loop.
- All heal feedback lands on stderr under the server's `guardStdout`, so the JSON-RPC stream on stdout stays clean.
- After the heal the machine is back up and containers came back fresh, so the gvproxy host port forwards re-registered (sites reachable on 443 again).

geodro:
Thanks for this, the shape is solid and the tests cover the heal states nicely, builds and passes clean here.

Two things I'd want to sort before it goes in. First the probe is unconditional, EnsureMachineResponsive runs on every StartFPM including on Linux where MachineHeal is nil and there's no machine VM that can stall or be healed, so it just adds a podman ps subprocess to every artisan and composer call for a check that can never do anything there. Can we gate it to darwin, either a runtime.GOOS check at the call site or an early return nil when MachineHeal is nil, so Linux pays nothing.

The bigger one, we already have the resume detection you'd want for this. Have a look at resumed() in internal/watcher/dns.go, it ticks every 30s and tells a real host wake from a merely delayed tick by comparing the wall clock gap against the monotonic gap, and on a resume it already fixes DNS and brings nginx back after sleep. That lives in the lerd-watcher daemon which is always up, and since the podman machine is one shared VM, healing it there fixes the MCP path too, proactively, once per wake, before the agent's next call ever lands. That's really the on wake probe you mention deferring in the description. I'd rather see the machine heal folded into that resumed() block as a darwin branch next to the existing linux nginx one, then the StartFPM probe can either drop out or stay as a cheap macOS only fallback. The reason for deferring, not spinning the VM on every laptop open, doesn't really bite here, resumed() only fires on a real suspend gap over 90s, and even then it only restarts if a bounded podman ps probe actually fails.

Happy to pair on the watcher wiring if that helps.

prabowosd:
Thanks for that, i will check and fix

prabowosd:
Both sorted, thanks for the steer, the watcher angle is clearly the right home for this.

**1. Linux pays nothing.** `EnsureMachineResponsive` now returns nil immediately when `MachineHeal` is nil, before the probe. So Linux runs no extra `podman ps` on artisan/composer (nothing there can stall or be healed anyway), and on macOS the StartFPM gate stays as a cheap reactive fallback. Updated the nil-hook test to assert the no-op skips the probe entirely.

**2. Proactive heal folded into `resumed()`.** Added a darwin `healMachine` dep, wired in `WatchDNS` next to the Linux nginx branch, and called from the same resume block in `tickDNS`. On a real wake (the existing >90s monotonic-vs-wall gap) it un-stalls the shared machine VM before the next agent call lands. It reuses `EnsureMachineResponsive`, so:
- it only restarts the VM when a bounded `podman ps` probe actually fails, no spin on a wake where the VM survived;
- the proactive and reactive paths share one `lastHealAt` cooldown, so they can't double stop-start;
- `isStopped` (config.IsStopped) keeps a deliberate `lerd stop` from resurrecting the machine.

On macOS the nginx deps are nil so its resume heal is a no-op, and on Linux `healMachine` is nil, so the two branches stay cleanly separated in the resume block. The StartFPM probe stays as the macOS-only fallback for the missed-resume edge you already document for nginx (watcher restarted during the outage) and for a stall that isn't a >90s suspend.

New tests: `TestTickDNS_resumeTriggersMachineHeal` (heals once on the resume tick, never on first/normal ticks) and `TestHealMachineOnResume` (no-op on unset hook / stopped stack). All `internal/podman` + `internal/watcher` + `internal/php` tests green.

prabowosd:
E2E'd the proactive watcher path too, against a real stopped machine.

Drove the actual resume tick (`tickDNS` → `resumed()` → `healMachineOnResume` → `defaultHealMachine` → `EnsureMachineResponsive`) with the machine stopped to stand in for a stalled post-suspend VM. The only seam is the clock (the same wall-vs-monotonic seam `resumed()` already takes for tests); `tickDNS`, the resume detection, the heal, and the VM restart are all real:

```
machine: stopped
  → tickDNS resume tick (fake-clock jump > 90s, monotonic frozen)
  → healMachineOnResume → defaultHealMachine → EnsureMachineResponsive
  → bounded `podman ps` probe fails → restart:
      "restarting the Podman Machine to remount it…"
      "Starting Podman Machine…"
      "Waiting for Podman Machine to be ready… ready"
  → re-probe OK
PASS (13.95s)
```

Confirmed the machine was up and containers came back after the tick, and a healthy machine takes the fast path (a second run with the VM already up is a 0.15s no-op, no restart) so it only spins the VM when the probe actually fails.

So both paths are now verified live: reactive (StartFPM gate on an MCP exec) and proactive (watcher resume tick). Ready for another look.

PR #736, opened 2026-07-03
release: v1.27.1

geodro:
lerd could only move a service's primary published port, so multi-port services like Mailpit, RustFS and Selenium had their secondary ports stuck on the defaults, and the dashboard link and iframe kept pointing at the old port whenever a port changed.

This adds a per-container-port override map so every published mapping is movable from the CLI with a --container flag on lerd service port, from MCP with container_port, and from the dashboard Ports modal, which now lists one editable host-port field per published port. The dashboard link follows whichever port its service is served on, and the port-ownership guard shifts secondary ports off host-owned defaults the same way it already did for the primary.


geodro:
Superseded by #737 on the correct hotfix/1.27.1 branch.

Issue #762, opened 2026-07-06
v1.27 review follow-ups: macOS compatibility


geodro:
Reviewing the v1.27 changes on macOS surfaced a few platform issues to fix.

The MCP dump tools and the TUI dumps panel dial the lerd-ui unix socket unconditionally, but that socket is only created on Linux, so on macOS every MCP dump, query, status and clear call and the TUI dumps panel fail to connect. They should use the TCP loopback client the CLI already switched to in v1.27.

The host-database port coexistence guard relies on a bind probe that only tests the loopback specifics, so a host server listening on all interfaces slips past it under SO_REUSEADDR on macOS, and lerd publishes its container onto the same port and collides. The probe should also test the all-interfaces wildcard.

The ssh-agent sidecar introduced the first bare named volume in the FPM template, and the launchd volume pre-create loop runs MkdirAll on the bare name, leaving a stray lerd-ssh-agent directory in the daemon working directory. Named volumes should be skipped.

The changelog entry for the host-database coexistence describes detection as keying off a data directory or live socket, which does not match the bind-probe implementation, so the wording needs correcting.

PR #763, opened 2026-07-06
fix: macOS compatibility follow-ups from the v1.27 review

geodro:
Reviewing the v1.27 changes on macOS turned up a handful of platform issues, all fixed here.

The MCP dump tools and the TUI dumps panel dialed the lerd-ui unix socket directly, but that socket is only bound on Linux, so on macOS they never connected. v1.27 moved the CLI dump client onto a TCP loopback transport for exactly this reason, and this routes the MCP and TUI callers through the same config.UIClientNetwork and config.UIClientAddr seam so the dump, query, status and clear surfaces work on macOS too.

The host-database coexistence guard decides whether to shift a service off its default port with a bind probe, but the probe only tried the loopback specifics. On macOS a specific-address bind slips past a wildcard or dual-stack listener under SO_REUSEADDR, so a host database listening on all interfaces read as free and lerd would publish its container onto the same port and collide with it. Bindable now also probes the all-interfaces wildcard, which catches a host server on all interfaces and, as a side effect, correctly sees a running container's own gvproxy-published port.

The lerd-ssh-agent sidecar added the first bare named volume to the FPM container template, and the launchd container writer pre-creates each volume source directory, so it ran MkdirAll on the bare name and left an empty lerd-ssh-agent directory in the daemon working directory. Named volumes are podman-managed and are now skipped, so only absolute bind-mount sources are created.

Finally the v1.27 changelog described the coexistence detection as keying off a populated data directory or a live socket, which never matched the bind-probe implementation, so the wording is corrected to describe the port-bindability probe.

Refs #762


prabowosd (commented):
Reviewed the macOS follow-ups, looks solid, nothing blocking. The OS-aware dial seam is applied consistently (MCP + TUI now match `cli/dump.go`), the wildcard bind probe's caller-side shift risks are already covered by the existing exclusions (`maybeShiftPublishedPort` only probes inactive services, `SetFPMPorts` excludes the version's own ports), and the launchd named-volume guard is sound. A few non-blocking nits:

1. `internal/freeport/freeport.go` — the new probe comment calls it "The IPv4 wildcard (0.0.0.0)", but Go binds a wildcard laddr on network `tcp` as a dual-stack `[::]` socket (as `freeport_test.go`'s own comment notes). Behaviour is fine, actually stronger since it also catches a v6-only wildcard listener; only the mechanism description is slightly off.

2. Stale comments now that the transport is OS-aware: `internal/mcp/dumps.go` still says "HTTP-over-Unix-socket helpers" / "over the local Unix socket", and `internal/tui/dumps.go`'s `runDumpsListener` doc still says "Unix-socket SSE endpoint". Worth the same wording fix this PR already applies to the changelog.

3. Question, not a blocker: `Bindable` runs a brief wildcard `listen()` on every quadlet write, could that trip the macOS application-firewall "accept incoming connections" prompt when ALF is enabled? ALF is off by default and the listener closes immediately, so likely fine; just flagging that it was considered.

Otherwise the tests are real behavioural tests and I could not find any missed callers.

geodro:
good catches on both, fixed the comment wordings in c6561823, the freeport one now describes the wildcard probe as the dual stack socket Go actually binds rather than an IPv4 only 0.0.0.0, and the mcp and tui dump docs no longer claim a unix socket. on the ALF question it should be fine, Bindable only ever binds and closes right away and never calls accept, and lerd already holds real host listeners like lerd-ui on 7073 plus the dump and access ports, so the momentary probe adds nothing new to the firewall surface. i confirmed the bind and close timing earlier while measuring the gvproxy release.

prabowosd (approved):
Thanks for the quick turnaround. The comment fixes in c6561823 read correctly now (the freeport probe is described as the dual-stack `[::]` socket Go actually binds, and the MCP/TUI dump docs no longer claim a Unix socket), and the ALF reasoning holds: `Bindable` only binds and closes, never calls `accept`, and lerd already owns real host listeners (lerd-ui on 7073, the dump/access ports), so the momentary probe adds nothing to the firewall surface. LGTM.

Issue #19, opened 2026-03-30
Getting an issue on php version


eznix86:
i did `laravel new` 

then i did `lerd link` it downloaded an image and all with php8.3 ( i think).

<img width="820" height="144" alt="Image" src="https://github.com/user-attachments/assets/e41e5b52-1a69-4cf4-90f7-edc395ac3f84" />

But i get this msg. So I ran `lerd use 8.4` and then `lerd unlink` then `lerd link` but i get: `Linked: lerd-test -> lerd-test.test (PHP 8.3, Node 22, Framework: laravel)`

So i think there is an issue somewhere. or I am doing something wrong here. Thanks

PR #769, opened 2026-07-06
feat(queue): derive queue:start command from the framework definition

prabowosd:
`lerd queue:start` (CLI) and the MCP `queue_start` tool both hardcoded `php artisan queue:work --queue/--tries/--timeout`, so a non-Laravel framework that defines a `queue` worker (CodeIgniter's `php spark queue:work`) got a broken command. The MCP path additionally built its own systemd unit, so it only worked on Linux.

### Change
- Add a `tune_command` template (with `{queue}`/`{tries}`/`{timeout}` placeholders) to `FrameworkWorker`; `renderQueueCommand` substitutes them so each framework declares its own flag syntax, falling back to the plain `command` when no template is set.
- Add `restart_command`; `QueueRestartForSite` runs it (no-op when a framework declares none, e.g. CodeIgniter) instead of hardcoding `php artisan queue:restart`. The legacy `Restart=on-failure` → `always` unit heal is preserved.
- Route the MCP `queue_start`/`queue_stop` tools through the shared cli lifecycle. `cli` injects `QueueStartTuned`/`QueueStopForSite` as func vars (it already imports `mcp`, so this avoids the `cli` → `mcp` → `cli` cycle), replacing the duplicate hardcoded systemd unit builder. The MCP path now derives the command from the framework and starts the worker cross-platform (systemd/launchd), like the web UI toggle.
- Remove the dead `buildQueueUnit`/`queueDependencyUnits` helpers (the live path is `WorkerStartForSite`) and their tests; add `TestRenderQueueCommand`.

### Verification
`go build`/`vet`/`go test ./internal/cli ./internal/mcp` green. End to end on a real CodeIgniter app:
- CLI: `lerd queue:start --queue emails --tries 5` runs `php spark queue:work emails -tries=5`.
- MCP: `queue_start` (queue=emails, tries=7) runs `php spark queue:work emails -tries=7` and succeeds on macOS (previously errored with "systemd DBus not available"); `queue_stop` stops it.
- Laravel still renders the artisan command.

The framework `tune_command`/`restart_command` values ship in lerd-env/frameworks#19.


geodro:
This is the right direction, moving the hardcoded artisan syntax into the framework definition, and the CodeIgniter case reads cleanly. Two small things before it goes in.

The renamed QueueStartTuned is exported now and its comment says it's shared by lerd queue:start and the MCP queue_start tool, but nothing in mcp actually calls it, and your own follow up note explains why it can't yet because of the cli to mcp import cycle. So right now it's an exported symbol with no external caller and a comment claiming a sharing that doesn't exist. Either keep it unexported as queueStartTuned, or drop the MCP mention from the comment until the primitive actually moves down a layer.

The other thing is docs. This changes user facing behaviour, queue:start working for non Laravel frameworks, but the PR only touches Go. Can you add a line to the relevant queue worker page under docs so the behaviour change is written down, the frameworks store PR covers the author side but the user side isn't documented anywhere yet.

prabowosd:
Both addressed in 45cd8fb9.

1. Unexported it to `queueStartTuned`. Heads-up on the timing: the follow-up commit dae1fe23 (pushed after your review) actually wires the MCP path now, `cli` injects `queueStartTuned`/`QueueStopForSite` into `mcp.QueueStartFn`/`QueueStopFn` before `mcp.Serve()`, which sidesteps the `cli` -> `mcp` -> `cli` cycle, so the MCP `queue_start` no longer builds its own hardcoded systemd unit and derives the command from the framework cross-platform. That said, the symbol is only ever referenced inside the `cli` package (`runQueueStart` and the hook registration), so it doesn't need to be exported, unexported and reworded the comment to describe the hook.

2. Added a line to `docs/usage/queue-workers.md` under the Queue worker section: it now works for CodeIgniter (`php spark queue:work` when `codeigniter4/queue` is installed) and notes that `--queue`/`--tries`/`--timeout` render into each framework's own flag syntax.

geodro (approved):

PR #808, opened 2026-07-08
release: v1.28.1

geodro:
Cuts v1.28.1. This patch gets the v1.28.0 max_allowed_packet change to actually reach running services on a plain lerd update rather than only on an explicit reinstall: reconcile now re-materialises a running service's preset config files and restarts the container when the config drifted since it last booted, covering both the default stack and store services. It also retries a transient store fetch, a slow raw.githubusercontent.com response used to drop a framework or service definition mid-refresh, so a timeout or 5xx now retries a few times with backoff while a 4xx still fails fast. Finally it refreshes the Go dependencies, moving x/net to v0.56.0 and re-tidying the module graph, which clears stale Dependabot alerts against already-patched versions.

The commit bumps the version to 1.28.1, adds the changelog section, and bumps the landing demo version. Full detail is in the changelog at https://lerd.sh/changelog.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -7,6 +7,21 @@ Lerd uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ---
 
+## [1.28.1] - 2026-07-08
+
+A patch release that gets the v1.28.0 `max_allowed_packet` change to actually reach running services on update, and hardens the framework and service store against transient fetch failures.
+
+### Fixed
+
+- **A shipped preset config change reaches a running service on update, not only on reinstall** (#805). The v1.28.0 bump of MySQL's `max_allowed_packet` only took effect after an explicit `lerd service reinstall`, because the reconcile pass regenerated a service's config files only when its unit was missing and never restarted a running container. Reconcile now re-materialises a running service's preset files and restarts the container when its config file is newer than the container's boot, meaning the config genuinely drifted, so a shipped preset config change lands on a plain `lerd update`. The mod time only advances on a real content change, so a steady state restarts nothing, and the same drift check covers both the default stack and store services.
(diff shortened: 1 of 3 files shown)


prabowosd:
Smoke-tested locally on macOS (nogui build). ✅ LGTM.

**Static gate**
- `gofmt -l .` clean, `go vet ./...` clean
- `go test -tags nogui ./...` → 3866 passed / 59 packages (incl. `serviceops/reconcile_test.go` + `store/store_test.go` covering #805/#807)
- `make install` succeeds; binary runs

**Behavioral — #805 (a shipped preset config reaches a running service on update)**

Had a real drift to test against: a running `mysql` service booted before the v1.28.0 bump, so its `max_allowed_packet` was still `64M` while the preset now ships `256M`.

| | mysql `max_allowed_packet` | mysql restarted | a steady custom service (mariadb) |
|--|--|--|--|
| before | 64M | — | running |
| after the reconcile pass (`install --from-update`, i.e. the `lerd update` tail) | **256M** | yes (to apply) | data intact, reachable |
| 2nd run (steady state) | 256M | **no** | **no** |

So the drifted service picked up the shipped config on the update path, and a second run restarted nothing — matches the "mod time only advances on a real content change, steady state restarts nothing" contract.

**Notes (not blockers)**
1. The first `install --from-update` also restarted mailpit/meilisearch/postgres/redis/rustfs (and the custom mariadb) citing "PublishPort changed". This was an artifact of swapping from a previously-installed release binary to a locally-built branch binary (ports re-derived), **not** #805 — the second run with the same binary restarted nothing. A real release→release update wouldn't hit it.
2. The DNS resolver step wanted sudo (`sudo: a terminal is required`) in the non-interactive path — non-fatal, the saved DNS config was preserved.
3. #807 (transient store-fetch retry) and #806 (x/net v0.56.0 + tidy) are covered by unit tests + `go.mod`; I didn't drive #807 at runtime (would need to simulate a raw.githubusercontent timeout).

PR #816, opened 2026-07-09
fix: stop stale service badges lingering on the site page

### LINKED ISSUE #815: Site SERVICES badge lingers for a service referenced only by a commented-out .env line
A site detail page enumerates its SERVICES badges from container hostname references found in the project `.env` (`envfile.ReferencesContainer`), matched against every default preset name.

The match scans the whole `.env` as one blob, so a disabled line still counts as a live reference. For example a site whose active DB is MariaDB:

```
#DB_HOST=lerd-mysql
DB_HOST=lerd-mariadb-10-11
```

keeps showing a MySQL badge. Because no MySQL container is running for it, the badge sits there permanently as a gray "Stopped" chip that never goes away, even though nothing on the site actually uses MySQL.

Expected: a commented-out reference should not add a service to the site SERVICES list.

Fix: iterate the `.env` line by line and skip commented-out lines (leading `#`) before the token match. An active, uncommented reference still wins when both a commented and an uncommented line mention the same service.

prabowosd:
A site detail page enumerates its SERVICES badges from container hostname references in the project `.env` (`envfile.ReferencesContainer`), matched against every default preset name. A badge could get stuck as a permanent gray "Stopped" chip that never went away. This fixes three ways that happened:

1. Commented-out references counted as live. The match scanned the whole `.env` as one blob, so a disabled line like `#DB_HOST=lerd-mysql` still counted even when the active `DB_HOST` pointed elsewhere. The `.env` is now scanned line by line and commented-out lines (leading `#`) are skipped. An active reference still wins if both a commented and an uncommented line mention the same service. The whole-token boundary check is unchanged, just moved into a per-line helper.
2. Inline comments on an otherwise live line. A trailing comment like `DB_HOST=lerd-mariadb # was lerd-mysql` still ghosted the commented service. A `#` preceded by whitespace is now stripped before the token match (dotenv convention), so a `#` inside a value (e.g. a password) is preserved.
3. Removed services still enumerated. `enrichServices` inferred default-preset services purely from `.env` references, with no install-state check, so a service whose quadlet had been removed kept ghosting on every site whose `.env` still pointed at it (e.g. a CodeIgniter site with `database.default.hostname=lerd-mysql` after mysql was removed). The `.env`-inferred default presets are now gated on `podman.QuadletInstalled`, matching the `ServiceInstalled` source of truth. An installed-but-stopped service still shows as "Stopped" (a useful signal); only a genuinely absent one is skipped.

Notes:

4. `ReferencesContainer` is also used by the unpause path to decide which services to start for a site, so this also stops unpause from starting a service that is only mentioned on a commented-out `.env` line, which is the intended behaviour.
5. `.lerd.yaml`-declared services are intentionally left ungated: they are an explicit declaration rather than something inferred from `.env`, so they show up whether or not the service happens to be installed.

Tests: `TestReferencesContainer` gains cases for commented, indented-comment, inline-comment, live-token-before-inline-comment, and hash-without-leading-space; a new `TestEnrichServices` case covers a referenced-but-uninstalled service being skipped while an installed one is still detected.

Closes #815

### DIFF
--- a/internal/envfile/envfile.go
+++ b/internal/envfile/envfile.go
@@ -202,11 +202,23 @@ func ReferencesContainer(content, serviceName string) bool {
 		if strings.HasPrefix(strings.TrimSpace(line), "#") {
 			continue
 		}
-		if lineReferencesNeedle(line, needle) {
+		if lineReferencesNeedle(stripInlineComment(line), needle) {
 			return true
 		}
 	}
 	return false
+}
+
+// stripInlineComment drops a trailing "# ..." comment from an .env line. A '#'
+// only starts a comment when preceded by whitespace (dotenv convention), so a
+// '#' inside a value (e.g. a password) is preserved.
+func stripInlineComment(line string) string {
+	for i := 1; i < len(line); i++ {
+		if line[i] == '#' && (line[i-1] == ' ' || line[i-1] == '\t') {
+			return line[:i]
+		}
+	}
+	return line
 }
 
 // lineReferencesNeedle reports whether line contains needle as a whole token
--- a/internal/envfile/envfile.go
+++ b/internal/envfile/envfile.go
@@ -193,16 +193,32 @@ func ReadValues(path string) map[string]string {
 
 // ReferencesContainer reports whether content references the lerd container
 // hostname "lerd-<serviceName>" as a whole token, so bare "postgres" is not
-// matched by a "lerd-postgres-18" reference (and vice versa).
+// matched by a "lerd-postgres-18" reference (and vice versa). Commented-out
+// lines are ignored so a disabled "#DB_HOST=lerd-mysql" doesn't keep a removed
+// service's badge alive on the site page.
 func ReferencesContainer(content, serviceName string) bool {
 	needle := "lerd-" + serviceName
+	for _, line := range strings.Split(content, "\n") {
+		if strings.HasPrefix(strings.TrimSpace(line), "#") {
+			continue
+		}
+		if lineReferencesNeedle(line, needle) {
+			return true
+		}
+	}
+	return false
+}
+
+// lineReferencesNeedle reports whether line contains needle as a whole token
+// (the next byte, if any, is not part of a service name).
+func lineReferencesNeedle(line, needle string) bool {
 	for i := 0; ; {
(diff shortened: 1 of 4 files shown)


geodro (dismissed):
nice catch on both, the commented line blob scan is a real bug and the removed service ghosting is the other half of it. ran it locally and the whole gate is green.

two things worth a mention. ReferencesContainer is shared with the unpause path, which uses it to work out which services to start for a site, so this also changes unpause to stop starting a service that is only mentioned on a commented line. that is the behaviour we want, but it is worth a line in the description since nothing covers it. and the skip is line level, so a trailing comment like DB_HOST=lerd-mariadb followed by a hash and was lerd-mysql on the same line still keeps the mysql chip. not asking you to chase that here, just noting the tests do not cover that shape.

gating on QuadletInstalled is the right call, it is exactly what ServiceInstalled already does for default presets, so it reuses the existing source of truth rather than inventing a second one. and leaving the .lerd.yaml declared services ungated is right too, those are an explicit declaration rather than something inferred from .env, so they should show up whether or not the service happens to be installed.

prabowosd:
Thanks for the review. Handled the inline-comment case in 9921012: a `#` preceded by whitespace is now stripped before the token match (dotenv convention), so `DB_HOST=lerd-mariadb # was lerd-mysql` no longer keeps the chip, while a `#` inside a value is preserved. Added `TestReferencesContainer` cases for inline comment, a live token before an inline comment, and a hash without a leading space.

Also added the unpause-path behaviour to the description as you suggested, and kept `.lerd.yaml`-declared services ungated per your note.

That new commit tripped the stale-review dismissal, so this needs a re-approve when you get a sec.

geodro (approved):
inline handling looks right, whitespace before the hash is the dotenv rule for unquoted values and the three new cases pin the boundaries nicely. ran it against every env file on my machine, nothing changes behaviour there, so this is a correctness fix rather than a live symptom, which is the good kind. one thing i noticed and am explicitly not asking you to fix, the strip does not know about quotes, so a value like DSN="# lerd-mysql" would lose its reference where real dotenv keeps the hash literal inside quotes. nobody writes that, and chasing it would mean a quote aware scanner for no gain. good to merge.

PR #834, opened 2026-07-10
fix: keep WebSocket upgrades out of the request-timing view

### LINKED ISSUE #831: Request timing records WebSocket upgrades as slow routes
nginx logs a proxied WebSocket once, when the connection closes, with status 101 and `$request_time` set to the whole lifetime of the socket. reqstats ingests that line as an ordinary request. The only filters on the ingest paths are the static-asset extension check and a zero request time, in `aggregator.go`, `idle_feed.go` and `store.go`, so a 101 goes straight through carrying a multi-thousand-second timing.

The absolute-slow rule then flags any route whose p95 clears a second from a single sample, which a closed socket does by three orders of magnitude. One WebSocket close is enough to plant a permanent fake slow route, drag the site median toward it, and fire the slow_route push notification.

Every site running Reverb has this. On my install the requests table holds 26 rows with status 101, all on `GET /app/{appKey}`, across two sites, the largest at 3,623,181 ms. Each one sits within about 25 seconds of a full hour, so these are healthy sockets reconnecting on a cycle rather than anything going wrong. The app key never collapses to `:id` either, since `isIDSegment` only folds digits, UUIDs and hex, so every key gets its own route row.

The log format already captures `$status` and nothing reads it. Dropping 101 at ingest is a universal nginx signal with no framework knowledge in it, so it belongs in reqstats rather than the store. The predicate is currently repeated in four places, which is the reason to give it a single home.

Worth deciding separately whether an upgraded connection should count as activity for idle-suspend, since the same datagram feeds that path.

geodro:
nginx logs a proxied WebSocket once, when the connection closes, with status 101 and the whole lifetime of the socket as its request time. reqstats treated that line as an ordinary request, so a single Reverb or Vite HMR connection was enough to plant a permanent slow route running thousands of times above the site baseline, drag the median it was being measured against, and fire a slow_route push. On my install every site running Reverb had one, the largest sitting at just over an hour.

The ingest filter now lives in one place, `reqstats.IsAppRequest`, which rejects an upgraded connection alongside the static assets and zero request times it already skipped. It had been copied into four call sites, the live aggregator, both read paths of the durable store, and the watcher fan-out, and they all go through the one predicate now, so every consumer agrees on what counts as a request the app actually served.

Filtering on read as well as on write means the upgrades already sitting in a store stop surfacing without needing a migration. Failed upgrade attempts keep counting, since a 502 from nginx while Reverb is down is a real request and belongs in the error rate.

The sites documentation now describes the exclusion.

Refs #831


prabowosd (approved):
LGTM. Root cause and fix both check out: nginx logs an upgraded socket once at close with status 101 and the whole connection lifetime as request_time, so a single long-lived Reverb/Vite HMR socket plants a permanent slow route. Unifying the four call sites behind IsAppRequest is the right move (verified: grep leaves only the definition, no inline IsStaticAsset || ms==0 remains anywhere). Filtering on read as well as write clears already-stored rows without a migration, and keeping non-101 statuses (502 while Reverb is down) preserves the error rate. Test coverage is thorough across aggregator, store, and the watcher fan-out; all green locally.

PR #835, opened 2026-07-10
fix: report the real p50 in the live request-timing snapshot

### LINKED ISSUE #832: route_timing always reports a p50_millis of 0
`RouteStat` carries `P50Millis` and `RecentP95Millis`, but the live aggregator's `snapshotLocked` only ever assigns `Route`, `Method`, `Example`, `P95Millis`, `Multiplier` and `Samples`. Both fields are populated only by the durable store's `SiteAnalytics`.

The MCP `diag route_timing` action and the UI endpoint behind it read the persisted live-aggregator snapshot through `LoadSite`, not `SiteAnalytics`, so `p50_millis` and `recent_p95_millis` come back as zero for every route on every call. An agent reading that output takes the zero at face value and concludes the route is normally instant, which is how this surfaced.

Either `snapshotLocked` should compute both from the ring it already has, or the fields should come off the struct that endpoint serializes.

geodro:
`RouteStat` carries a median and a recent p95, but the live aggregator only ever filled in the p95, the multiplier and the sample count. Both other fields went out as zero for every route, and that snapshot is exactly what `diag route_timing`, `diag optimize_route` and `site_stats` serve. A caller reading a p50 of zero concluded the route was normally instant, when in truth nothing had been measured. That is not hypothetical, it is how an assistant misread a WebSocket outlier on one of my sites and built a whole diagnosis on top of it.

The snapshot now takes the median from the same recent window it already reads for the p95. That window is bounded to the last few minutes by construction, so a route's p95 over it is its recent p95, and the field carries it rather than a zero that would understate a route still running slow. The durable store is unaffected, since its window spans the whole selected range and it computes both figures already.

No documentation change: the sites page already describes a table of every route with its p50 and p95, so this brings the code in line with what was written.

Refs #832


prabowosd (dismissed):
LGTM. Filling P50Millis and RecentP95Millis from the same recent window the p95 already reads is correct: the live aggregator's window is bounded to recentWindow by construction, so its p95 is the recent p95, and a zero p50 previously read to a caller as a normally-instant route. RecentP95Millis equaling P95Millis in the live path is an expected consequence of that (the durable store is where the two legitimately diverge over a longer range), not a bug. Test asserts p50 sits below the tail on a heavy-tailed route. Green locally.

PR #840, opened 2026-07-10
fix: install ext-ftp in the PHP images so BundledExtensions is honest

### LINKED ISSUE #837: PHP image is missing ext-ftp, and BundledExtensions claims it is bundled
Magento requires `ext-ftp`. `composer install` on a Magento project fails the platform check until it is passed `--ignore-platform-req=ext-ftp`, and `php -m` in the FPM container confirms `ftp` is absent.

Separately, `internal/podman/extensions.go` lists `ftp` under "always-compiled PHP core". That is not true of our image, so `lerd park` treats it as available and never warns about it.

Add `ftp` to the `docker-php-ext-install` list in the php-fpm Containerfile, which also makes the existing `extensions.go` entry honest. This needs a base image rebuild and publish to take effect on installs.

prabowosd:
ext-ftp was neither compiled into the FPM base nor added via `docker-php-ext-install`, yet `internal/podman/extensions.go` listed it under "always-compiled PHP core". Two things followed from that. `lerd park` read the extension as available and never warned, and a Magento `composer install` failed its platform check until it was passed `--ignore-platform-req=ext-ftp`; `php -m` in the FPM container confirms `ftp` is absent.

This adds `ftp` to the FPM Containerfile's `docker-php-ext-install` list, and to the FrankenPHP runtime extension set so the documented FPM/Octane parity holds rather than the Octane image quietly diverging. The `extensions.go` entry moves out of the always-compiled core group into the `docker-php-ext-install` group, where it now truthfully belongs, so the set `park` checks against stays accurate.

`ftp` compiles on the alpine base with no new apk dependency: OpenSSL is already present for its optional ftps support, and `docker-php-ext-install ftp` builds fine without it. It needs a base image rebuild and publish to take effect on installs.

Refs #837


geodro:
nice catch, this is byte for byte what I ended up with independently, including moving ftp out of the always compiled group and adding it to frankenPHPRuntimeExtensions. I checked it against a real magento 2.4.9 install and composer check-platform-reqs now reports ext-ftp success with zero failures, so the fix itself is right.

two things before it goes in though. there is no test, and the invariant that actually broke here is worth guarding, a check in internal/podman that every extension the containerfile installs is present in BundledExtensions would have caught it. the parity test already has an fpmDockerExtInstallList parser you can reuse instead of writing another one. second, docs/usage/php.md still lists the old extension set, ftp needs adding there.

prabowosd:
Both addressed in the latest commit.

- **Test**: `TestBundledExtensionsCoverContainerfile` in `internal/podman/extensions_test.go` reuses `fpmDockerExtInstallList` and asserts every extension the FPM Containerfile installs is present in `BundledExtensions`, so the two can no longer drift.
- **Docs**: added `ftp` to the bundled-extension list in `docs/usage/php.md`.

And nice, glad the fix matched yours byte for byte, and thanks for the real magento check-platform-reqs confirmation.

geodro (changes requested):
the diagnosis is right and I checked it against my install, php -m has no ftp, extension_loaded('ftp') is false, and the base image configure line has no enable-ftp, so BundledExtensions really was advertising something that was never in the image. the containerfile line lands in the docker-php-ext-install list too, the hunk header showing apk add is just git picking a stale context line. that part I would take as is.

the regression test does not test the regression though. TestBundledExtensionsCoverContainerfile asserts that everything the containerfile installs appears in BundledExtensions, but the bug was the opposite, BundledExtensions claimed ftp and nothing installed it. I put the bug back exactly, ftp out of the containerfile and back into the always compiled core group, and the new test passes, and so does the rest of the podman package. the comment says it guards the invariant that broke in 837, and it guards a different one. because the old containerfile never mentioned ftp, the loop never looks at it.

what is worth pinning is that every name in BundledExtensions is actually available. you cannot get that from the containerfile alone because of the core group, so the cheap version is to make the core group a named constant and assert that BundledExtensions equals core plus the parsed install list. that breaks the moment ftp moves between the groups or leaves the containerfile, which is exactly what 837 was.

while I had the bug restored, with ftp gone from the containerfile but still sitting in frankenPHPRuntimeExtensions, the frankenphp parity test stayed green as well, so nothing currently stops the octane image from claiming an extension the fpm image does not build, which is the divergence the body says this protects against.

one more thing that fell out of checking. I diffed the whole BundledExtensions list against what actually loads in lerd-php84-fpm and got two names back, ftp and opcache. opcache is not quite a false positive, extension_loaded('opcache') is false in the container and composer publishes it as ext-zend-opcache, so a project requiring ext-opcache matches the literal opcache string in the bundle, park stays quiet, and composer install fails its platform check anyway. same symptom as 837 with a different cause. not for this pr, I will open a separate issue.

two housekeeping things, the body has a verified locally paragraph which we do not put in prs, and it says Closes #837, which reads as a bug report, so it should be Refs #837 and stay open until the stable release ships.

prabowosd:
You are right, and thanks for restoring the bug to prove it. My test guards the reverse direction of the one that broke, so it passes with ftp back in the core group and nothing installing it. That was the wrong invariant.

I want to pin what you described (every name in BundledExtensions is actually available), and I have one question on the exact shape before I push a second wrong test.

BundledExtensions is three groups: the always-compiled core, the docker-php-ext-install list, and the PECL set (redis, imagick, igbinary, mongodb, pcov, xdebug). `fpmDockerExtInstallList` only parses the docker-php-ext-install block, so `core + parsed install list` is missing the PECL names. Two ways to make the assertion whole:

1. Make all three groups named constants (`coreExtensions`, `dockerInstallExtensions`, `peclExtensions`), build `BundledExtensions` from them, and assert `dockerInstallExtensions` equals the parsed containerfile install list (and, if worth it, add a PECL parser and assert `peclExtensions` equals the parsed `pecl install` / git-clone-enabled set). That breaks the instant ftp leaves the containerfile or moves between the install and core constants.
2. Keep it minimal: named `coreExtensions` constant, extend the parser to also collect the `pecl install` / `docker-php-ext-enable` names, and assert `BundledExtensions == coreExtensions + parsedContainerfileExtensions` as sets.

I lean towards 1 since it also gives the frankenphp parity test a real list to check against, closing the divergence you noted where the octane image can claim an extension the fpm image never builds. One caveat either way: a core-group name that is not actually always-compiled still can only be falsified by `php -m`, which is what the base-images smoke test already does. Which shape do you want?

Body is updated: dropped the verified-locally paragraph and switched to Refs #837.

geodro:
neither shape catches 837, and I would rather you knew that before building one.

BundledExtensions is built from the same constants the assertion would compare against, so core plus install equals BundledExtensions is true by construction. ftp lived in the core group. under option 1, dockerInstallExtensions equals the parsed containerfile still holds because neither side has ftp, and coreExtensions is a free constant that nothing falsifies. under option 2 the equality is a tautology for the same reason. the core group is the whole problem, it is an unverifiable claim about what upstream php fpm alpine compiles in, and nothing in our own containerfile can contradict it.

the base-images smoke test also does not do what you think. it checks that the module count is at least 50, then spot checks eight names, curl gd intl mbstring pdo_mysql redis xdebug zip. it never compares against BundledExtensions. that is precisely why ftp sailed through it for as long as it did.

so the real guard belongs there. make base-images compare php -m against BundledExtensions exactly and fail when the image does not report a name we advertise. that catches 837, it catches the next one, and it would have caught the opcache naming problem I opened as 842, since php -m prints Zend OPcache and never opcache.

for this pr I would keep the unit test modest and honest. extend the parser to also pick up the pecl install and docker-php-ext-enable names, assert that everything the containerfile installs appears in BundledExtensions, and say in the comment that this direction catches an extension added to the image and never listed, not the 837 direction. that is a real invariant, it is just not the one the test name claimed.

on frankenphp you do not need the constants either. TestFrankenPHPRuntimeExtensionParity asserts fpm is a subset of franken plus builtins and never checks the reverse, which is why my restored bug stayed green with ftp sitting in the franken list and gone from the containerfile. add the other direction, every name in frankenPHPRuntimeExtensions appears in the containerfile install list or the pecl set, and that divergence closes with no refactor at all.

so, not one or two. keep a parser based subset test here, add the reverse franken check, and put the actual 837 guard in base-images.yml where php -m already runs. happy to split the ci change into its own issue if you would rather land this pr on the fix plus the honest test.

thanks for updating the body.

prabowosd:
done, and split as you offered.

- the unit test is honest now. it says what it actually catches, an extension added to the image and never listed, and its comment points at #856 for the real #837 direction. the parser (`fpmContainerfileExtensions`) now also collects the `docker-php-ext-enable` names, so opcache and the pecl set are covered, not just the `docker-php-ext-install` block. spx and lerd_devtools stay out: they are lerd-internal best-effort tooling, not project extensions.
- added the reverse to `TestFrankenPHPRuntimeExtensionParity`: every `frankenPHPRuntimeExtensions` name must actually be built by the FPM containerfile. one note, opcache lives on a `docker-php-ext-enable` line, not the install block or pecl, so the reverse check reads the enable names too rather than "install list or pecl set", otherwise opcache would false-fail. restored your bug locally (ftp back in the core group, out of the containerfile) and the base-images.yml php -m guard is the only thing that catches it, which is #856.
- the real #837 guard is #856: compare `php -m` in the built image against BundledExtensions in base-images.yml. i left the naming caveat there (Zend OPcache vs opcache, #842) since it needs the same apples-to-apples handling.
- merged main (the branch was behind).

full suite green.

geodro (approved):

PR #841, opened 2026-07-10
feat(worker-heal): flag a live worker whose server has died as unreachable

### LINKED ISSUE #839: Worker health treats a live process as a healthy worker, so a dead vite dev server reports as running
workerheal only looks for units systemd has put in the failed state, and the State field comment already reserves room for others. A host worker that keeps its process alive after its server has died is therefore invisible to it, and lerd worker list, the dashboard, and the TUI all report it as running.

This turned up on a real site. Editing .env made Laravel's vite plugin restart the dev server, the restart ran a codegen step that failed while the FPM container happened to be cycling, and the exception aborted the restart before the HTTP server rebound. The node process survived underneath fnm exec and npm, so systemd kept the unit active and running. Vite held no listening socket at all, while public/hot still advertised the port from before the restart, so every asset request the browser made was refused while lerd insisted the worker was healthy. The failure is silent and the reported state is the opposite of the truth, which is the part worth fixing.

The probe should ask whether the worker's server is reachable rather than whether its process exists. That has to stay framework-agnostic, so how to find the port belongs in the framework store yaml beside the worker, not in Go. The worker schema already carries a check block, so a sibling health block could declare where the port is published, for instance a file the dev server writes on boot. When the process is alive and nothing is accepting on that port, the worker reports unhealthy rather than running, and heal restarts it. Workers that declare no port keep today's behaviour exactly.

prabowosd:
Worker health only looked at systemd's failed state, so a worker that keeps its process alive after its server has died (a vite dev server that crashes under npm while the node process lingers) stayed `active`, and lerd reported it as running while nothing was listening. The reported state was the opposite of the truth and the failure was silent.

A worker can now declare a `health` block naming a file its server writes on boot (vite's `public/hot`, which holds the dev-server URL). While the unit is active, `Detect` reads that file and makes a short TCP dial to its host and port. Process-alive-but-unreachable is reported as a new `unreachable` state, and `HealAll` restarts the unit (a plain start is a no-op on a running unit) rather than starting it. Where the port lives stays in the framework store, never in Go, so it is framework-agnostic. A worker with no `health` block keeps the process-only liveness check. A missing `url_file` is treated as not-probeable, so idle-suspend clearing `public/hot` never trips a false alarm.

The framework resolves once per site inside `Detect` (never per worker), backed by an mtime-cached `composer.json` parse so `DetectMajorVersion` stops reparsing on a steady tick and `Detect`'s hot-endpoint contract stays honest. `Unreachable` is a distinct `WorkerInfo` state, not folded into `Failing` (which keeps meaning systemd failed): the dashboard and TUI worker rows render it as its own state with its own badge, and `Enrich` gives it a dedicated not-accepting line instead of reading the journal off an active unit.

The companion `health` block for the vite worker ships in a paired lerd-env/frameworks PR.

Per-worktree worker heal and gating the probe on `ActiveEnterTimestamp` (plus the grace rule) are split into #853 as a fast-follow, since both need the batched unit-state cache extended to carry `ActiveEnterTimestamp` + `WorkingDirectory`, the one source that feeds both in a single call per tick.

Refs #839

### DIFF
--- a/docs/usage/framework-workers.md
+++ b/docs/usage/framework-workers.md
@@ -48,6 +48,19 @@ workers:
 
 Port assignment scans all proxy port env keys across all sites to prevent collisions between different workers and frameworks.
 
+**Server health probe**: A worker whose process can outlive its server (a Vite dev server that dies under `npm` while the Node process lingers) declares a `health` block, so lerd probes reachability rather than mere process liveness:
+
+```yaml
+workers:
+  vite:
+    command: npm run dev
+    host: true
+    health:
+      url_file: public/hot   # a file the server writes on boot, holding its URL
+```
+
+`url_file` names a file the dev server writes when it binds (Vite's `public/hot` holds a URL like `http://[::1]:5173`). While the process is up, lerd reads that file and makes a short TCP dial to its host and port; if nothing is accepting, the worker reports **unreachable** instead of running and [worker-heal](/usage/worker-heal) restarts it. A worker with no `health` block keeps the process-only liveness check. A missing `url_file` is treated as not-probeable, so idle-suspend clearing `public/hot` never trips a false alarm.
+
 **Host workers**: Workers that need to run on the host instead of inside the PHP-FPM container set `host: true`. The command runs via fnm at the project's pinned Node.js version. This is used for tools like Vite that need direct filesystem access for HMR:
 
 ```yaml
--- a/docs/usage/worker-heal.md
+++ b/docs/usage/worker-heal.md
@@ -13,6 +13,12 @@ Two common paths:
 
 The second case requires fixing the underlying error before heal will stick. Heal will start the unit, but if it crashes again it'll burn through the rate limit and re-enter `failed`. The dashboard banner will reappear; check the worker logs (`lerd worker logs <name>` or the dashboard logs pane) for the real cause.
 
+## Workers that run but stop serving
+
+A worker can keep its process alive after its server has died: a Vite dev server whose HTTP server crashes under `npm` while the Node process lingers. systemd still reports the unit active, so process liveness alone reads it as running when nothing is listening.
+
+A worker that declares a [`health` block](/usage/framework-workers) is probed for reachability instead. While its process is up, lerd reads the URL file the server writes on boot (Vite's `public/hot`) and dials its host and port; if nothing is accepting, the worker is reported **unreachable** and heal restarts it (a plain start is a no-op on a still-running unit). Workers with no `health` block keep the process-only check.
+
 ## CLI
 
 ```sh
--- a/internal/config/framework.go
+++ b/internal/config/framework.go
@@ -148,6 +148,7 @@ type FrameworkWorker struct {
 	ExcludeCheck   *FrameworkRule `yaml:"exclude_check,omitempty"`  // only show when check FAILS (e.g. queue is hidden when laravel/horizon is installed because horizon supersedes it)
 	ConflictsWith  []string       `yaml:"conflicts_with,omitempty"` // workers to stop before starting this one (e.g. horizon conflicts_with queue)
 	Proxy          *WorkerProxy   `yaml:"proxy,omitempty"`          // WebSocket/HTTP proxy config for nginx
+	Health         *WorkerHealth  `yaml:"health,omitempty"`         // reachability probe: process alive but server not accepting = unhealthy
 	Host           bool           `yaml:"host,omitempty"`           // run on the host via fnm instead of inside the PHP-FPM container
 	// PerWorktree opts the worker into running independently per git worktree
 	// (lerd-<wname>-<site>-<wt>). Defaults to false; set true on workers that
@@ -177,6 +178,17 @@ type WorkerProxy struct {
 	Path        string `yaml:"path"`                   // URL path to proxy (e.g. "/app")
 	PortEnvKey  string `yaml:"port_env_key,omitempty"` // env key holding the port (e.g. "REVERB_SERVER_PORT")
 	DefaultPort int    `yaml:"default_port,omitempty"` // fallback port if env key is missing (default: 8080)
+}
+
+// WorkerHealth declares how to tell whether a worker's server is actually
+// reachable, not merely that its process is alive. A dev server (vite under
+// fnm/npm) can keep its process up after its HTTP server has died, so systemd
+// still reports the unit active while nothing is listening. Where the port lives
+// is declared here, never in Go: URLFile names a file the server writes on boot
+// (vite's public/hot) whose contents carry the URL to probe. A worker with no
+// Health block keeps the process-only liveness check unchanged.
+type WorkerHealth struct {
+	URLFile string `yaml:"url_file,omitempty"` // file (relative to site root) holding the server URL, e.g. "public/hot"
 }
 
 // FrameworkLogSource describes where application log files live for a framework.
--- a/internal/config/framework.go
+++ b/internal/config/framework.go
@@ -1910,14 +1910,54 @@ func DetectMajorVersion(projectDir, frameworkName string) string {
 	return ""
 }
 
-func detectVersionFromComposer(projectDir string, rules []FrameworkRule) string {
-	data, err := os.ReadFile(filepath.Join(projectDir, "composer.json"))
-	if err != nil {
-		return ""
+// composerParseCache memoises the parsed top-level sections of a project's
+// composer.json, keyed by path and invalidated by mtime+size. DetectMajorVersion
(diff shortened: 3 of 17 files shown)


geodro (changes requested):
the probe design is right, and the not probeable versus unreachable split via the probed return is exactly the shape this needed. restarting rather than starting an unreachable unit is also correct, since a plain start is a no op on a live unit. two things I want changed before this goes in.

defaultWorkerReachable calls GetFrameworkForDir from inside the per unit loop in Detect, so it runs once for every active worker, and lerd-ui runs Detect every five seconds. that is not a cpu problem. I measured it, a refused dial on loopback is about four microseconds and parsing composer.json is about twelve, so the whole thing is a rounding error. the problem is that GetFrameworkForDir is not just a lookup. when the store yaml is older than twenty four hours it calls frameworkFetchHook, and autoFetchFramework has no throttle, no backoff and no negative caching. a failed fetch leaves the file mtime untouched, so shouldFetch is true again on the next call. offline, or on a framework the store has no definition for, that becomes a network attempt per active worker every five seconds. before this pr Detect never touched the framework store at all.

the fix is small. Detect already builds meta per site, so resolve the Framework into it once per site, lazily, and look the worker up from that. the comment saying the framework read is cached and lands once per site is describing the code you want rather than the code that is there. loadFrameworkYAML and LoadProjectConfig are mtime cached, but DetectMajorVersion rereads and reparses composer.json on every call.

second, vite is per_worktree true, and Detect cannot see worktree units at all. workerUnitName gives a worktree the form lerd-vite-ws-feat-x, as TestWorkerUnitName_worktreePath pins, so after the lerd prefix comes off Detect looks for a site suffix on vite-ws-feat-x, never matches ws, and skips the unit. meanwhile enrichWorktreeWorkers in this same pr does probe the worktree path. so a worktree whose vite has died shows as failing on the dashboard while lerd worker heal does not know it exists. the skip predates this pr, but this pr is what makes it matter, because the worker the feature is built around is the one that runs per worktree.

third, siteinfo folds unreachable into Failing, and WorkerInfo has no field for the new state, so the dashboard and tui will say failed for a unit systemd calls active, with no journal line behind it to explain why. you added a real distinction in workerheal and then dropped it at the display layer. that may be deliberate, but it should be a decision rather than a side effect.

smaller, at a five second cadence a single transient probe failure reaches queueWorkerFailureNotifications and fires a push. in practice the laravel vite plugin removes public/hot when the server closes, and a missing file is correctly treated as not probeable, so the window is narrow, but gating the dial on the hot file mtime being older than the unit ActiveEnterTimestamp would close it and cost nothing. also worth knowing that enrichWorkers dials on every snapshot while Detect dials every five seconds, so there are two dials per vite per cycle. harmless, just noting it.

two housekeeping things, the body has a verified locally paragraph which we do not put in prs, and it says Closes #839, but that one is labelled bug so it should be Refs #839 and stay open until the stable release ships.

prabowosd:
Thanks, this is a thorough read. Agreed on all of it.

1. **Framework resolve in the loop** — you are right, the comment describes the code I wanted, not what is there. I will resolve the Framework once per site into the `meta` map, lazily, and look the worker up from that, so Detect touches the store at most once per site and never per worker. Good catch that the real cost is `autoFetchFramework` having no backoff or negative cache, not the dial.

2. **Worktree units** — I will teach Detect to resolve the worktree unit form (`lerd-<worker>-<site>-<wt>`) so heal covers the per-worktree vite too, since that is the worker this feature is actually about. It is the right time to fix the skip.

3. **Display state** — deliberate call to make, not a side effect, fair. My preference is to add an `Unreachable` field to `WorkerInfo` and render it as its own state in the dashboard and TUI, so an active-but-unreachable unit does not read as `failed` with no journal line behind it, keeping the distinction Detect already draws. Happy to go that way unless you would rather fold it and label the row explicitly some other way. Which do you prefer?

4. **Transient probe / push** — I will gate the dial on the hot-file mtime being older than the unit `ActiveEnterTimestamp`, so a just-restarted vite that has not rewritten `public/hot` yet is never dialed. Noted on the two dials per cycle; I will keep it in mind but leave it unless you want them coalesced.

5. **Housekeeping** — will drop the verified-locally paragraph and switch to `Refs #839` so it stays open until the stable release.

I will push the first, second, fourth and fifth now and hold the third for your call.

geodro:
go with the Unreachable field, and it is not merely cosmetic, there is a concrete bug sitting behind that choice. Enrich only skips the journal lookup for expected-but-stopped, with a comment explaining that a cleanly stopped worker's last line reads as a false error. an unreachable unit is active, so Enrich will call readLastError on it and attach whatever the live vite last logged, which on my box right now is a wayfinder plugin info line. the dashboard would render a row labelled failed with a cheerful info message presented as its cause. so add the field, render it as its own state, and give Enrich the same skip, or better a dedicated line like server not accepting on the port in public/hot. HumanState already falls through to the raw string, so unreachable renders without touching it.

keep Failing meaning systemd failed. no new card, no new banner, the existing worker row badge with its own icon is enough.

on the mtime gate, it is safe for the case we actually hit, and I want to spell out why because it is not obvious. our incident was an in process restart. vite printed .env changed, restarting server and never exited, so ActiveEnterTimestamp never moved, the unit still reported active running since 09:57:58 nearly two hours after the server died at 11:18:48. public/hot was written at the first bind just after 09:57:58, so its mtime is newer than ActiveEnterTimestamp and your gate still dials. good.

one thing to think about before you build it. the plugin removes hot on exit, it registers clean on process exit and converts SIGTERM, SIGINT and SIGHUP into process.exit, so a systemd restart normally leaves no hot file behind. that means a vite whose process survives a restart but never binds has no hot at all, and absent is treated as not probeable, so it would be invisible forever. I have not reproduced that, so treat it as a question rather than a finding, but if it is reachable then the rule that closes it is the mirror of your gate, a unit active for longer than some grace period with no url_file present is also unreachable. worth settling now, since the gate and the grace rule read the same timestamp.

last thing. Detect's doc comment promises no per site .lerd.yaml or composer.json reads, no extra subprocess calls, safe to invoke from a hot endpoint, and mcp does call it from one. resolving once per site still reads composer.json once per site per tick through DetectMajorVersion, so either that comment changes with the code or the resolve needs a cache with its own invalidation. worth deciding which, rather than leaving the comment describing the old contract.

you are also BEHIND main now, worth merging before you push the rest.

prabowosd:
pushed the first batch addressing the review.

- framework now resolves once per site inside Detect, never per worker, backed by an mtime-cached composer.json parse (`parseComposerJSON`, same shape as `loadFrameworkYAML`) so `DetectMajorVersion` stops reparsing on a steady tick and Detect's hot-endpoint contract stays honest. doc comment updated to match.
- Enrich no longer reads the journal for an unreachable unit. it is active, so the last line is whatever the live server logged and reads as a false cause; it gets a dedicated not-accepting line instead, framework-agnostic (no `public/hot` hardcoded in Go).
- `WorkerInfo` gets a distinct `Unreachable` field, set by the probe instead of `Failing`, rendered as its own state (tui glyph + web amber ringed dot). `Failing` keeps meaning systemd failed, no new card or banner.

two left, and both are a heads-up before I wire them so I don't go the wrong way:

1. **worktree units** — I'll teach Detect the `lerd-<worker>-<site>-<wt>` form and read the worktree path from the unit's `WorkingDirectory` (a file read, not a git subprocess), scoped to active per_worktree health units, so it stays inside the no-subprocess contract.
2. **mtime gate + grace rule** — both need `ActiveEnterTimestamp`, which nothing reads today. I'll extend the batched unit-state cache to capture it in the same call it already makes, so there's no new per-unit dial and the gate and the grace rule read one source. grace = active longer than the gate window with no `url_file` present, the mirror of the gate, to close the never-binds-no-hot case you flagged.

shout if you'd rather either come off a different source before I build it.

geodro (approved):
looks good, the live but dead detection and the display split are cleanly scoped and the suite is green. happy to land this batch now and pick up worktree heal and the timestamp gate as a fast follow off a fresh issue, with the batched cache extended to carry ActiveEnterTimestamp and WorkingDirectory so both features read from one call per tick.

prabowosd:
split it, per the call. this pr is now just the live-but-dead flag + the display split (`Unreachable` as its own state, `Failing` still = systemd failed). the first batch above covers the framework-resolve-once + composer cache, the Enrich line, and the `WorkerInfo.Unreachable` render.

per-worktree heal + the `ActiveEnterTimestamp` gate/grace are #853, with the batched unit-state cache extended to carry `ActiveEnterTimestamp` + `WorkingDirectory` as the single source that feeds both in one call per tick. updated the pr body to match the narrower scope.

Issue #846, opened 2026-07-10
The site Env tab hardcodes a root .env instead of resolving the framework's dotenv file


geodro:
`siteHasEnv` in `internal/ui/server.go` stats a literal `.env` in the project root, and `has_env` gates the Env tab in `SiteDetail.svelte`. It never asks the framework where its env file lives, although `fw.Env.Resolve(path)` already answers exactly that and is what the CLI and the doctor use.

For frameworks whose dotenv file is not a root `.env`, that is wrong in two ways. CakePHP reads `config/.env`, so it has never had an Env tab at all, and deep-linking to `#sites/<domain>/env`, which the doctor's edit-env action does, bounces back to Overview. Symfony reads `.env.local` but ships a committed root `.env` next to it, so the literal stat passes by accident and the tab shows and edits the wrong file.

Three hardcoded assumptions cause this and all three should be driven by what the framework declares through `fw.Env.Resolve`. `siteHasEnv` stats the wrong path. `listEnvFiles` scans only the project root, so a dotenv file in a subdirectory is never listed. And `envFileRe`, `^\.env(\.[A-Za-z][A-Za-z0-9_-]*)?$`, rejects any name containing a slash, so even a hand-built request for `config/.env` is refused. Resolving the framework's declared file and allowing its real location through all three fixes CakePHP and points Symfony at `.env.local`.

Scope is dotenv files only. The Env tab earns its place for dotenv because it is a flat key=value file that pairs with service-preset proposals and inline edits. Frameworks whose env lives in a PHP file, WordPress `wp-config.php`, Magento `app/etc/env.php`, and Drupal's `web/sites/default/settings.php` fallback, are full PHP source that developers already edit in their IDE, so a UI editor adds little there and risks mangling syntactically significant code. Those stay out of scope and keep their current behaviour of no Env tab. This change only makes the tab find and edit a real dotenv file wherever the framework puts it.

hiSandog:
The framework-specific dotenv path should probably be resolved through the same configuration source used by the runtime rather than maintained as a second mapping in the UI. How should the tab handle frameworks that load multiple files or apply environment-specific precedence such as `.env.local`? A test with a non-root dotenv file and a missing optional override would clarify both display order and fallback behavior.

geodro:
Yes, that was the whole point of the fix, it resolves through fw.Env.Resolve and GetFrameworkForDir, the same version aware store definition the CLI, the doctor and the service wiring already use, so there is no second mapping living in the UI.

[...]

PR #858, opened 2026-07-11
fix(ui): resolve the framework's dotenv file for the site Env tab

prabowosd:
The site Env tab decided everything from a literal root .env, never asking the framework where its dotenv actually lives, even though fw.Env.Resolve already answers that for the CLI and the doctor. That broke any framework whose env file isn't a root .env. CakePHP reads config/.env, so it never got an Env tab at all and the doctor's edit-env deep-link to #sites/<domain>/env bounced back to Overview. Symfony ships a committed root .env next to its real .env.local, so the literal stat passed by accident and the tab edited the wrong file.

This routes all three hardcoded spots through the framework's declared file. siteHasEnv now stats the resolved file, so a subdirectory dotenv surfaces the tab and its deep-link lands. listEnvFiles picks up a dotenv that lives outside the project root and lists the framework's file first. envFileFromQuery lets that declared file through even when it carries a slash the old regex rejected, while every other name still has to be a root dotenv variant. The five env endpoints now share one resolver.

The tab stays dotenv-only on purpose. Frameworks whose config lives in PHP source, WordPress wp-config.php, Magento app/etc/env.php, Drupal's settings.php, resolve to a non-dotenv format and keep their current behaviour of no tab, since a flat key=value editor would risk mangling code that matters.

The Symfony built-in now declares .env.local with a .env fallback so the tab edits the local-override file instead of the committed one.

Refs #846


hiSandog:
One useful regression case may be a framework whose resolved dotenv path contains multiple directory components, such as `config/environments/.env.local`. It could verify that the declared path is accepted and listed first, while an arbitrary query path with the same shape is still rejected. A second case where the preferred file is absent but the framework fallback exists would also clarify the intended fallback behavior.

prabowosd:
Good call, added both. \`TestSiteEnv_frameworkNestedDotenvAcceptVsReject\` uses \`config/environments/.env.local\` and asserts the declared path is accepted and listed first while an arbitrary path of the same nested shape (\`config/environments/.env.other\`) is still rejected, so the whitelist is keyed on the exact declared file, not any slashed name. \`TestSiteEnv_frameworkFallbackWhenPrimaryMissing\` covers a project with only the fallback (no primary), confirming the tab still resolves. Both green.

geodro (changes requested):
The backend resolution here is right, and collapsing the five endpoints onto one resolver is a nice cleanup, but I put it on my install and drove it against a real Symfony site and there are two things to fix before this can go in.

The first is that the UI side never got updated, so the Symfony fix doesn't actually reach the browser. envQS in sites.ts drops the file param whenever the value is .env, and SiteEnvTab hardcodes .env as both its initial selection and its snap back when the selected file disappears. My symfony.test has both .env and .env.local, so the file list now comes back with .env.local first, the tab opens with .env selected, sends no file param, and the server correctly resolves and returns .env.local. I checksummed it, the editor is showing .env.local content under a dropdown that says .env. And because .env is exactly the value envQS strips, picking .env in the dropdown can never load the committed file, it just re requests the framework default. For CakePHP the list is only config/.env, the initial .env is not in it, and the snap back sets it to .env again, so the select sits on a value that isn't one of its options. The propose banner also stops appearing, since it is gated on file matching the proposed file and that is now the resolved path. The fix is small, default the selection to empty or to files[0] since the server already sorts the primary first, always send the file param when it is set, and snap back to files[0] instead of .env.

The second is that backup and restore break for exactly the subdirectory case this PR is meant to enable. envCfgFile passes the whole relative path as BkpName, so for config/.env cfgedit writes the backup into config/.env.bkp.timestamp while ListBackups scans the project root and matches basenames against a regex anchored on config/.env, which can never match. I proved it with a throwaway test, the save reports a backup name, the file lands in config/, ListBackups returns nothing, and ReadBackup on the name the save just handed back says the file does not exist. So on a CakePHP site the backup badge never shows, restore is impossible, and backups quietly accumulate inside the committed config directory. Splitting it, BkpDir as the dir of the joined path and BkpName as the base name, sorts it.

Two smaller things while you're in there. The declared path is trusted verbatim, no Clean or IsLocal check, so a framework yaml declaring something like ../../.ssh/config becomes readable and writable through the env endpoints. The store can already run commands so it isn't a new trust boundary, but the guard is one line and worth having. And /env/files still lists root dotenv variants for a non dotenv framework like WordPress or Magento while read, write and backups now 400 for them, which is invisible today only because the tab is hidden, but the endpoints should agree.

Everything else checks out, has_env behaves correctly across all sixteen of my sites, Laravel true, WordPress and Magento false, Symfony true, frameworkless sites still fall back to a root .env, and the full gate is green.

Thanks @hiSandog for the nested path case, that was a good catch and the test that came out of it is the right shape. If you're enjoying lerd, a star on the repo helps a lot.

prabowosd:
Thanks @geodro, good catches, all four addressed in 0e1e0cc.

**UI not reaching the browser.** `envQS` no longer drops the `file` param when it equals `.env`, so picking `.env` in the dropdown loads the committed file instead of re-requesting the framework default. `SiteEnvTab` now starts with an empty selection and snaps to `files[0]` (the primary the server already sorts first) rather than a hardcoded `.env`, both on init and when the selected file disappears. So the dropdown label and the editor contents agree, and the propose banner reappears for a CakePHP-style site whose only file is `config/.env`.

**Backup/restore for the subdirectory case.** `envCfgFile` now derives `BkpDir` from the joined path's dir and `BkpName` from its base, so a nested `config/.env` backs up next to the file and stays listable and restorable. A root `.env` resolves to the same `BkpDir`/`BkpName` as before, so no change there.

**Declared path trusted verbatim.** Added the one-line `filepath.IsLocal` guard in `frameworkEnvFile`, so an absolute or `../..` declaration (`../../.ssh/config`) is rejected and the framework simply gets no editable dotenv.

**/env/files disagreeing with the other endpoints.** `listEnvFiles` now returns nothing for a non-dotenv framework (WordPress, Magento) where read, write, and backup already 400, so the endpoints agree.

Regression tests cover the nested-backup round-trip (save, list, read), the traversal rejection, and the empty listing for a non-dotenv framework. Full gate is green on my install: `go build`, `go vet`, `gofmt`, and the `internal/ui` + `internal/cfgedit` suites pass, and `svelte-check` is clean on the two files touched.

geodro:
nice, the shape of this is right, one resolver shared by all five endpoints and a traversal guard on the declared path.

two things though. frameworkEnvFile goes through config.GetFramework, but the env wiring and the doctor, the code this cites as already knowing the answer, go through config.GetFrameworkForDir, and those two return different definitions for any built in framework. GetFramework hands back the Go built in and never looks at the versioned store yaml, GetFrameworkForDir detects the version from composer.lock and returns symfony@8 from the store. on main today GetFramework("symfony") gives env.file .env while GetFrameworkForDir gives .env.local, so the Env tab would resolve against a definition nothing else uses, and the per version store defs (symfony 7 vs 8, cakephp 4 vs 5) get ignored entirely. resolveEnvTarget already has dir in hand, so it can call GetFrameworkForDir(site.Framework, dir) instead.

second, the symfony change belongs in the store rather than in Go. frameworks/symfony/7.yaml and 8.yaml already declare file: .env.local, what they're missing is the fallback, so adding fallback_file: .env and fallback_format: dotenv there is the whole fix, and it reaches every install within a day with no release. as it stands the built in gets the fallback but the doctor and the service wiring read the store, so a symfony project carrying only a committed .env ends up with the UI editing .env while the doctor and the wiring point at a .env.local that isn't there, which is the same split this is trying to close.

worth a docs pass too, the Env tab appearing for cakephp and symfony editing .env.local are both user visible.

prabowosd:
Thanks @geodro, both addressed in 1abdc37.

**Env tab resolving against a definition nothing else uses.** `frameworkEnvFile` now goes through `GetFrameworkForDir(site.Framework, dir)` instead of `GetFramework`, so the tab resolves against the same version-aware store definition the env wiring and the doctor use, and the per-version defs (Symfony 7 vs 8, CakePHP 4 vs 5) are respected rather than ignored.

**Symfony change belongs in the store.** Reverted the built-in `symfony` env change back to `.env`, and moved the fix into the store defs: lerd-env/frameworks#25 adds `fallback_file: .env` + `fallback_format: dotenv` to `symfony/7.yaml` and `8.yaml`. That reaches every install within a day with no release, and keeps the built-in, the doctor, and the wiring pointed at the same file, so the split you described (UI editing `.env` while the doctor points at a missing `.env.local`) closes.

Added a regression test (`TestSiteEnv_frameworkResolvesVersionedStoreDef`) asserting the versioned store def wins over the built-in, and did the docs pass in `docs/features/env-setup.md` covering the framework-aware file selection (Symfony `.env.local` with `.env` fallback, CakePHP `config/.env`, no tab for non-dotenv frameworks). `go build`, `go vet`, `gofmt`, and the `internal/ui` + `internal/config` suites are green.

geodro (approved):
thanks for sticking with this through three rounds, the shape you landed on is right and the backend resolves the same file everywhere now, one resolver shared by the endpoints was the correct call.

I pushed a commit on top rather than sending you round four. propose was the one route still going through GetFramework, so on symfony it proposed against .env while the tab had .env.local open, and since the banner is gated on those two agreeing the whole missing keys flow was invisible. listEnvFiles only rescued the declared file when it carried a slash, so a framework declaring a root file the regex does not match got a visible tab with an empty dropdown and the editor silently fell back to a .env the framework never reads. on the front end the store helpers still defaulted to .env, which after your envQS change meant every call that omitted the file forced the committed one instead of letting the server resolve it, and that is also what turned the vitest suite red, worth running make test-ui as well as svelte-check, the latter does not cover it. the tab also fetched on its empty initial selection, loading one file and then immediately reloading the one it snapped to.

I also tried a symlink containment guard on the declared path and then took it back out. it broke projects that share one .env through a symlink, has_env went false and every env route started refusing, and it did not close anything a hostile repo could not already do without it, so your lexical guard stays as it was.

drove it on a real symfony site, view, save with backup, and restore, on both .env and .env.local, everything round trips clean and the files stay scoped to their own backups.

one thing before this lands, it waits on lerd-env/frameworks#25. the symfony store defs declare .env.local with no fallback today, so a symfony project carrying only a committed .env comes back has_env false and loses a tab it has on main. I reproduced that on my install and confirmed your fallback closes it, so the store goes first and this follows.

PR #936, opened 2026-07-15
fix(dns): keep .test resolving offline on systemd-resolved systems

### LINKED ISSUE #914: Laravel Site gets auto refresh in ~30 sec interval.
I am running a Laravel Livewire ( v13) site with Vite. The page keeps refreshing in 30s interval. 
I have been able to fix this using LLM, but this have to done on every restart. 

Let me know what things i have to share for this issue.

My Machine:
╭──────────────────┬────────────────────────────────────────────────────────╮
│ kernel-name      │ Linux                                                  │
│ nodename         │ cachyos-kushal                                         │
│ kernel-release   │ 7.1.3-2-cachyos                                        │
│ kernel-version   │ #1 SMP PREEMPT_DYNAMIC Wed, 08 Jul 2026 18:34:01 +0000 │
│ machine          │ x86_64                                                 │
│ operating-system │ GNU/Linux                                              │
╰──────────────────┴────────────────────────────────────────────────────────╯

### LINKED ISSUE #916: .test names stop resolving after a reboot on NetworkManager plus systemd-resolved hosts
On hosts where NetworkManager manages systemd-resolved, lerd routes .test through an NM dispatcher script that applies per-interface DNS via resolvectl on each interface up event. The script runs at boot and does apply the routing, but NetworkManager re-pushes its own DNS to resolved a fraction of a second later and wipes the per-interface entry, so .test is unresolvable after a reboot until lerd is run again by hand. This shows up on Fedora and its atomic images, where NM's final DNS update lands after the dispatcher hook. Ubuntu happens not to hit the race, but the fix should cover every NM host.

lerd will also write a global systemd-resolved drop-in for the routing domain. The global scope is not per-interface, so NM leaves it alone and .test resolves straight after a boot. lerd currently deletes that drop-in whenever NM is present, on the assumption that NM overrides it, which is not true for a routing domain, so that deletion is replaced with writing it. The dispatcher stays in place to keep the upstream DNS in sync when the network changes. The drop-in will honour a configured custom tld rather than a hardcoded .test.

### LINKED ISSUE #933: .test domains not accessible when offline (no internet)
## Environment

OS: Linux Mint 22.3 Zena Ubuntu Noble-based
Network Manager: NetworkManager 1.46.0 nmcli

## Describe the problem

When the machine is offline not connected to the internet, .test domains served by lerd are not accessible in the browser. DNS resolution for .test domains appears to fail when there is no active internet connection.

Expected behavior: .test domains should resolve locally regardless of internet connectivity, since they are local development domains handled by lerd DNS configuration.

## Steps to reproduce

1. Start a site with lerd e.g. app.test
2. Confirm the site is accessible while online
3. Disconnect from the internet disable Wi-Fi or unplug ethernet
4. Try to access app.test in the browser
5. Page fails to load DNS resolution fails

## Additional context

This suggests that the local DNS configuration for .test domains may depend on upstream DNS servers or internet connectivity rather than being fully resolved locally.

geodro:
systemd-resolved refuses to resolve anything once no link is routable, over glibc NSS as well as resolvectl, and that holds for a loopback DNS server whether it is declared per interface or globally. A fresh `.test` lookup therefore failed with no network at all, even though lerd-dns kept answering on `127.0.0.1:5300`. The giveaway was a site that kept working while the browser stayed open on a cached answer but failed as soon as it was reopened.

lerd0 is an always-up dummy link carrying the `~test` route, so resolved keeps forwarding `.test` to lerd-dns with no network at all. lerd owns it through `lerd-dns-link.service` and tells NetworkManager to leave it alone. An NM connection would show up as a togglable network in the desktop network menu, and switching it off breaks offline `.test` with no symptom until the user next loses the network, which is the worst shape a bug can have. Unmanaged also stops NM re-pushing DNS over the link, which removes the reboot race entirely rather than working around it. resolved only grants a link a DNS scope once it carries a routable address, so lerd0 takes `192.0.2.1/32` from the range RFC 5737 reserves for documentation: it never appears on a real network, so the host route it installs cannot shadow something the user needs to reach. Hosts running a pre-release build get migrated off the old keyfile connection automatically.

Both resolved paths need the link. The NetworkManager-less case, Arch and omarchy, declared the resolver in a global drop-in instead. That was insufficient, because resolved refuses a global loopback server offline just the same, and harmful, because a global server is a catch-all: every ordinary name went to lerd-dns, and offline dnsmasq forwarded it to an upstream that was not there, hanging each lookup for twenty seconds. lerd0 carries the same route scoped to `~test`, so that drop-in is now removed rather than written.

The dispatcher no longer regenerates the dnsmasq address records. They are lerd policy, loopback normally and the host's LAN IP under `lan:expose`, and only the Go side knows which applies. Rewriting them from a template on every interface up event clobbered `lan:expose` back to loopback and dropped the AAAA record. Losing that record is what left dnsmasq forwarding `.test` AAAA queries to the upstream, and once that upstream went away it cost twenty seconds on every offline lookup, measured through NSS rather than the five seconds resolvectl reports.

Fallback DNS servers are turned off alongside the link. Making resolved willing to serve `.test` offline is the same flag that makes it willing to chase names it cannot reach, so it works through its fallback servers one at a time and every offline lookup of an ordinary name hangs. This is not something the link can dodge; the two behaviours are one flag. Debian, Ubuntu and Fedora already ship `FallbackDNS` empty, which is exactly why it never showed there, so this takes effect only on Arch and its derivatives and brings them in line. The cost is that a broken upstream DNS now fails instead of quietly routing queries to a public resolver, and uninstalling puts the fallbacks back.

`lerd start` now refreshes the sudoers drop-in before reapplying DNS config. A release that adds a privileged step ships new grants, and a host that upgraded the binary without re-running install would otherwise reach `ConfigureResolver` without them, where the watcher has no one to answer a password prompt. Disabling DNS also tears the resolver plumbing down on Linux now, mirroring what macOS already did, so a disabled setup no longer leaves a dispatcher and an interface pointing at a dnsmasq that has been removed.

Refs #933
Refs #916
Refs #914


geodro:
@aijorgenson can you test this one on nixos?

geodro:
@prabowosd can you test this one on macos?

prabowosd:
Yess ill check

prabowosd (dismissed):
Tested on macOS (Darwin 25.5.0), nogui build path. All green.

**Local gate**
- `gofmt -l .` clean, `go vet -tags nogui ./...` clean
- `go test -tags nogui ./internal/dns/... ./internal/cli/...` → 1230 passed
- `make install` OK, new build installed, lerd-ui + watcher restarted

**macOS `.test` resolution still correct**

This host runs `lan:expose` (`exposed: true`), so the LAN IP is the expected answer:

```
<site>.test → <lan-ipv4>     (A)
<site>.test → <lan-ipv6>     (AAAA)
```

That matches the PR's stated policy: loopback normally, host LAN IP under `lan:expose`.

**Why this is safe on macOS**

The diff is Linux internals. In `diagnose.go` the macOS branch is only a constant rename (`macOSKind = "macOS native dnsmasq"`, same value), and the new "offline .test route" rung is gated behind `runtime.GOOS == "linux"`. No behavior change on the mac path.

**Non-blocking, pre-existing (not from this PR)**

`lerd dns:check` reports a false negative on a `lan:expose` host: the dnsmasq A record answers the LAN IP, but the check hardcodes `want 127.0.0.1` and is not `lan:expose`-aware, so it prints "DNS is NOT working" even though resolution is correct. It also emits a `systemctl --user` (Linux) hint on macOS. Both are untouched by this PR. Worth a separate follow-up if you want, does not block this.

LGTM on the macOS side.

PR #1009, opened 2026-07-20
fix(pest): fetch Playwright browsers without its Node extractor

### LINKED ISSUE #1006: `lerd pest:browser install` hangs forever with no output — Playwright's Node unzip extractor deadlocks mid-write inside the FPM container
## Summary

`lerd pest:browser install` downloads the Playwright browser archives successfully (progress bar reaches 100%), then hangs indefinitely with **no further output, no error, and no timeout**. It never reaches the "shimming to musl chromium" step, so `lerd pest:browser doctor` permanently reports:

```
[x] Playwright browser shimmed to musl chromium
      → lerd pest:browser install
```

Root cause: the `playwright install chromium` invocation that lerd runs inside the FPM container uses Playwright's **Node-based zip extractor**, which deadlocks partway through writing files into `/root/.cache/ms-playwright` (the bind mount to `~/.local/share/lerd/playwright`). Extraction stops mid-file and the process spins forever.

The container's own `/usr/bin/unzip` extracts the **exact same archive to the exact same destination in ~5–8 seconds**, so this is not a slow-mount or disk-space problem.

A secondary problem makes this much harder to recover from: pressing `Ctrl+C` kills only the host-side `lerd` process. The in-container Node processes survive and keep holding Playwright's `__dirlock`, so **every subsequent `lerd pest:browser install` blocks on that stale lock, also with zero output**. Users end up in a loop where the command appears to hang instantly and forever.

## Environment

| | |
|---|---|
| lerd | 1.29.0 (commit `c738e930ef3f0fc98e16e2fe251b25927f5a8fe2`, built 2026-07-15) |
| Host | macOS 27.0 (build 26A5378j), Apple Silicon (arm64) |
| podman | 6.0.1, `podman-machine-default`, applehv, 5 CPUs / 4 GiB RAM / 100 GiB disk |
| Container | `lerd-php84-fpm` — Alpine 3.24.1, aarch64, PHP 8.4.23 |
| Node (in container) | v24.17.0 |
| Playwright | `@playwright/test` 1.56.1 / `playwright-core` 1.56.1 |
| Pest | `pestphp/pest` ^4.0, `pestphp/pest-plugin-browser` ^4.3 |

Browser cache mount (from `podman inspect lerd-php84-fpm`):

```
/Users/<user>/.local/share/lerd/playwright -> /root/.cache/ms-playwright (bind)
```

## Steps to reproduce

1. Fresh project, `pestphp/pest-plugin-browser` in `composer.json`, `playwright` in `node_modules`.
2. Run `lerd pest:browser install`.

## Actual behaviour

```
Setting up Pest browser testing for PHP 8.4...
  Chromium 150.0.7871.128 Alpine Linux
Downloading the Playwright browser registry...
Removing unused browser at /root/.cache/ms-playwright/chromium-1194
BEWARE: your OS is not officially supported by Playwright; downloading fallback build for ubuntu20.04-arm64.
Downloading Chromium 141.0.7390.37 (playwright build v1194) from https://cdn.playwright.dev/...
|■■■■...■■■■| 100% of 175.9 MiB
                                     ← hangs here forever (waited 24+ minutes)
```

## Expected behaviour

Extraction completes, lerd shims the browser binary to the image's musl `chromium`, and `lerd pest:browser doctor` reports all checks green.

### 1. The download is fine; the extraction deadlocks

The archive is fully downloaded and intact in the container:

```console
$ podman exec lerd-php84-fpm ls -la /tmp/playwright-download-cljegb/
-rw-r--r-- 1 root root 184483730  playwright-download-chromium-ubuntu20.04-arm64-1194.zip
```

But the destination contains only **2 of 474 files**, and the second one is **truncated**:

```console
$ podman exec lerd-php84-fpm ls -la /root/.cache/ms-playwright/chromium-1194/chrome-linux/
-rwxr-xr-x 1 root root 201625  libEGL.so      ← zip says 414496 bytes; truncated mid-write
-rwxr-xr-x 1 root root  37394  xdg-mime
```

Zip entry order confirms it stalls while writing entry #2:

```console
$ podman exec lerd-php84-fpm unzip -l <zip> | head
        0   chrome-linux/
    37394   chrome-linux/xdg-mime
   414496   chrome-linux/libEGL.so     ← extraction dies here
        0   chrome-linux/locales/
...
608969378   474 files
```

The extractor process stays alive indefinitely, pinned at that point:

```console
$ podman exec lerd-php84-fpm ps aux | grep -i playwright
  9 root  0:02 node ./node_modules/.bin/playwright install chromium
 20 root  0:02 /usr/bin/node .../playwright-core/lib/server/registry/oopDownloadBrowserMain.js
```

Zero byte growth over repeated 20-second samples. Not slow — deadlocked.

**This is deterministic**: three separate runs produced byte-identical partial output (`libEGL.so` at exactly 201625 bytes).

### 2. It is not a resource problem

```console
$ podman exec lerd-php84-fpm df -h /
overlay  99.4G  8.0G  91.3G   8% /
$ podman exec lerd-php84-fpm free -m
Mem: 3890 total, 2085 available
```

Plenty of disk and memory.

### 3. It is not the bind mount / virtiofs being slow

Plain `unzip` from the same container, same source zip, same destination path on the same bind mount:

```console
$ podman exec lerd-php84-fpm unzip -q -o <zip> -d /root/.cache/ms-playwright/chromium-1194
real  0m8.27s
user  0m4.49s
sys   0m1.93s
```

**8 seconds for all 474 files / 609 MB.** So the mount handles heavy write throughput fine — it is specifically Playwright's Node extractor that wedges.

### 4. Every downloaded component hits the same wall

Not chromium-specific. After manually extracting chromium, re-running `lerd pest:browser install` progressed to the next component and deadlocked identically:

| Component | Result |
|---|---|
| `chromium-1194` | hung at 2/474 files |
| `ffmpeg-1011` | hung with `ffmpeg-linux` at 3350668 bytes (zip says 3367792 — truncated) |
| `chromium_headless_shell-1194` | hung at 1/12 files |

So this reproduces on a 1.7 MB archive as well as a 175 MB one — size is not the trigger.

### 5. `Ctrl+C` leaves orphans that make retries hang silently

When the user aborts the apparently-stuck command, `lerd` dies on the host but the in-container Node processes survive and continue holding `/root/.cache/ms-playwright/__dirlock`. Every subsequent run then blocks on the lock:

```console
$ podman exec lerd-php84-fpm ps aux | grep playwright
  9 root  node ./node_modules/.bin/playwright install chromium     ← orphan from aborted run
 20 root  node .../oopDownloadBrowserMain.js                        ← still holding __dirlock
 53 root  node ./node_modules/.bin/playwright install chromium      ← new run, blocked on lock
```

From the user's perspective the command now hangs **immediately** after `Downloading the Playwright browser registry...` with no message at all — indistinguishable from the original failure, and with no hint that a lock is involved.

## Workaround (what fixed it)

Bypass Playwright's Node extractor entirely and use the `unzip` binary already present in the FPM image:

```bash
# 1. Kill orphaned in-container extractors (note: do NOT let pkill match its own cmdline)
podman exec lerd-php84-fpm sh -c 'pkill -f oopDownloadBrowserMain; pkill -f "playwright inst"'

# 2. Clear the stale lock
podman exec lerd-php84-fpm rm -rf /root/.cache/ms-playwright/__dirlock

# 3. Extract each already-downloaded archive by hand
podman exec lerd-php84-fpm sh -c '
  D=/root/.cache/ms-playwright/chromium-1194
  rm -rf "$D"; mkdir -p "$D"
  unzip -q -o /tmp/playwright-download-*/playwright-download-chromium-*-1194.zip -d "$D"
  chmod +x "$D"/chrome-linux/chrome "$D"/chrome-linux/chrome_sandbox "$D"/chrome-linux/chrome_crashpad_handler
  touch "$D/INSTALLATION_COMPLETE"'
# ...repeat for ffmpeg-1011 and chromium_headless_shell-1194

# 4. Re-run — Playwright now sees the browsers as installed and skips the download
lerd pest:browser install
```

Result:

```
Setting up Pest browser testing for PHP 8.4...
  Chromium 150.0.7871.128 Alpine Linux
Downloading the Playwright browser registry...
  shimmed 1 browser binary(ies) to system chromium

Pest browser testing is ready for PHP 8.4.
```

```
$ lerd pest:browser doctor
Pest browser testing — PHP 8.4:
  [✓] pestphp/pest-plugin-browser in composer.json
  [✓] chromium baked into the FPM image
  [✓] PHP 8.4 FPM container is running
  [✓] chromium present in the container
  [✓] playwright npm package installed
  [✓] Playwright browser shimmed to musl chromium
  [✓] Playwright server boots (run-server, launchServer mode)
```

The `INSTALLATION_COMPLETE` marker file per browser directory is what makes Playwright's registry accept a manually-extracted browser.

## Suggested fixes

1. **Don't rely on Playwright's Node extractor.** Since lerd shims the binary to the image's musl `chromium` anyway, the downloaded glibc build only needs to exist on disk in the right shape. Downloading the archive and extracting it with the image's `unzip` (already present) is both far faster and avoids this deadlock entirely.

2. **Add a stall timeout.** Any step that can hang forever should abort with a diagnostic after N seconds of zero progress, rather than sitting silent indefinitely.

3. **Detect and clear stale `__dirlock`.** Before invoking `playwright install`, check for a `__dirlock` with no live owning process and remove it — or at minimum print `waiting for Playwright's install lock held by PID …` so the user knows what is happening.

4. **Clean up in-container children on `Ctrl+C`.** Trap the signal on the host side and kill the corresponding `podman exec` child processes, so an abort leaves a recoverable state instead of a wedged one.

5. **Surface extraction progress.** The jump from `100% of 175.9 MiB` straight to silence gives no signal that the failure is in extraction rather than download. A single `Extracting …` line would have made this self-diagnosing.

## Possibly related upstream

The deadlock appears to be in Playwright's Node extraction path (`oopDownloadBrowserMain.js`) under Node v24 on Alpine/musl over a virtiofs-backed bind mount. I have not isolated which of those factors is decisive — extracting to a container-local (overlay) path instead of the bind mount would be a useful next triage step, and would also indicate whether relocating `PLAYWRIGHT_BROWSERS_PATH` off the bind mount is a viable fix on lerd's side.

## Leftovers worth noting

Each aborted attempt leaves its full archive behind in the container under `/tmp/playwright-download-*`. After four attempts this accumulated **459 MB** of orphaned zips that nothing cleans up.

khanalpride:
`lerd pest:browser install` downloaded every Playwright archive to 100 percent and then hung indefinitely with no output, no error and no timeout, so the shim step never ran and the doctor permanently reported the browser as unshimmed. The stall is Playwright's Node zip extractor deadlocking partway through writing into the bind-mounted browser cache in the FPM container. It is deterministic, reproduces on a 1.7 MB archive as readily as a 175 MB one, and the container's own unzip writes the same archive to the same destination in about eight seconds.

The install now asks the project's own Playwright where each component belongs and what to download, via a dry run, then fetches the archives with curl and unpacks them with the image's unzip, marking each directory installed the way the registry expects. Playwright still decides the components and revisions, so the cache layout stays correct for whatever version the project pins. A stalled download aborts with an error rather than hanging, and each component now says what it is downloading and extracting.

Aborting made things worse before this: Ctrl+C killed the host side only, leaving the in-container installer alive holding Playwright's install lock, so every retry blocked on that lock with no output at all and looked identical to the original hang. The lock and any orphaned downloader are now cleared before the install starts, after a failure, and on interrupt, and the half-downloaded archives that nothing ever reclaimed are reaped too.

Refs #1006


khanalpride:
reference for this one is https://github.com/lerd-env/lerd/issues/1006, the full diagnosis with the partial extraction evidence and the stale dirlock loop is there. verified the new fetch path inside lerd-php84-fpm against playwright-core 1.56.1, all three components downloaded and extracted in 36 seconds, 467 files, and a following playwright install chromium accepted the cache and downloaded nothing.

geodro:
nice fix, tested it end to end on a clean php 8.4 setup with the cache on the host bind mount, so the same condition as the deadlock. install came in around 36 seconds, three components, and a following playwright install chromium accepted the cache and downloaded nothing. re runs are idempotent, doctor reports the browser shimmed, and interrupting mid download left no lock, no orphan curl and no marker, with the retry picking up just the missing component. really solid.

one thing before merge, the empty plan branch execs into playwright's own installer, which is the path that deadlocks, with no timeout. i forced it twice in the container, once with dry run output the awk couldn't parse and once with a dry run that exits non zero, and both hung past twenty seconds. the second is sneakier since plan capture takes its status from awk, so set -e never fires and playwright's real error gets swallowed.

capturing the dry run on its own line, failing loudly if it errors, then erroring out when the plan is empty, covers both, failing fast beats falling back to something known to hang here.

not a regression either way, on main that installer is the only path so it hangs every time, this just narrows it.

PR #1105, opened 2026-07-23
fix(podman): isolate custom services from the host /etc/hosts

erhanurgun:
## Problem

Custom services get no explicit `/etc/hosts` mount, so podman derives the file from `base_hosts_file`, which defaults to the host's own `/etc/hosts`. Whatever a developer keeps there is inherited by every service container, and a line mapping a service name to loopback shadows the container name aardvark-dns would otherwise resolve.

The sidecars are where this shows, since phpMyAdmin, pgAdmin and Mongo Express all connect by container name:

```
mysqli::real_connect(): (HY000/2002): Connection refused
```

PHP-FPM containers already bind-mount a lerd-managed hosts file, so the two paths were inconsistent. The mount does not supply these services anything they read, it exists to cut the host inheritance.

## Fix

Mount the managed hosts file in the `else` branch of `GenerateCustomQuadlet`, so every custom service overrides host `/etc/hosts` inheritance. `share_hosts` services keep the browser-hosts variant for `.test` resolution. This covers every service the generator produces, not only the three admin dashboards.

The mount needs its source to exist first. Podman creates a directory at a missing `Volume=` source, and on macOS `precreateBindMountDirs` runs off `WriteQuadletDiff` and would `MkdirAll` the hosts path itself, after which every `WriteContainerHosts` fails with `is a directory` and only warns. `lerd start` hits that ordering because `reconcileCustomServices` runs before `WriteContainerHosts`, and `lerd service install` never touches hosts at all. So `EnsureCustomServiceQuadlet` now pre-creates the mount source before generating the quadlet, covering the `share_hosts` variant too.

The pre-create stays podman-free and writes only the two localhost lines. The real entries come from `WriteContainerHosts` at start and link time, and a service resolves its peers through container DNS rather than this file. `ensureFPMHostsFile` no longer treats a file without the host gateway line as finished, so an FPM quadlet written after a service still gets `host.containers.internal`.


geodro:
thanks for the patch, can I ask what the mount is actually for? I can't work out what these sidecars read out of that file, it carries localhost, host.containers.internal and the .test domains, and phpMyAdmin, pgAdmin and Mongo Express want none of those, since lerd-phpmyadmin resolves lerd-mysql through aardvark either way. So it reads like it's there to shadow the inherited file rather than supply anything, and I want to be sure that's the intent in case I'm missing a case.

The inheritance is real, I checked and my sidecars are carrying my host /etc/hosts wholesale. What I can't place is the 127.0.0.1 lerd-mysql line, I grepped the repo including install.sh and nothing on our side writes it, and the client shim runs in a throwaway container on the lerd network so it resolves through container DNS without touching the host file. Any chance you remember what put it there? If lerd wrote it that's the bug to fix, and if not we're really deciding whether to isolate sidecars on principle, which is fair but a different change.

If it is just isolation we want, podman's --hosts-file none does that while keeping its own generated entries including the container's own name, though it was renamed across versions so it'd need a capability check like the StopTimeout one.

One heads up either way, WriteFPMQuadlet calls ensureFPMHostsFile first because podman creates a directory at a missing Volume source, and EnsureCustomServiceQuadlet doesn't materialise the hosts file, so this has every custom service depending on a path nothing guarantees exists.

erhanurgun:
you are right on both counts, the mount supplies nothing to those sidecars, it only stops them inheriting the host file, and I should have framed it that way. lerd does not write that line either, I grepped the same places you did and then went back through every tag from 1.0.0 to 1.28.1, nothing on our side has ever touched the host /etc/hosts, the only sudo writes are the resolver, sudoers and the NetworkManager drop ins, and client-exec runs its throwaway container without touching it exactly as you said. the line came from my own machine, added by hand at some point so something running on the host could reach lerd-mysql by name, so the older lerd attribution in the description was wrong and I have rewritten it around isolation.

on --hosts-file none, that flag is podman 5 only, 4.9.3 here does not have it and neither does the 4.9.3 man page, only --add-host and --no-hosts, so a capability check would still need the mount as the 4.x arm, and version.go already treats 4.9.3 as first class for exactly that reason with StopTimeout. --no-hosts does avoid the bind mount question entirely but it hands the localhost line back to whatever the image ships, which trades one unknown for another.

your last point is the real one and it is worse than a missing guarantee. EnsureCustomServiceQuadlet never materialises the file, and on macOS precreateBindMountDirs runs off WriteQuadletDiff and will MkdirAll the hosts path, so the quadlet write itself creates the directory and the WriteContainerHosts that follows fails with is a directory and only warns. lerd start hits that ordering, reconcileCustomServices runs before WriteContainerHosts, and lerd service install never touches hosts at all. 1.15.1 fixed this class for the three FPM sources and the changelog says the scan covered service quadlets, which held then because they had no hosts mount, this PR breaks that assumption. the ensure step now runs inside EnsureCustomServiceQuadlet before the generate call, it stays podman free and writes only the two localhost lines since the real entries come from WriteContainerHosts at start and link time, and ensureFPMHostsFile no longer treats a file without the gateway line as finished so an FPM quadlet written after a service still gets host.containers.internal.

three things next to it that I left out of this PR, the share_hosts path has the same gap and predates this change since browser-hosts is never materialised, ensureFPMHostsFile returned early whenever hosts existed so it never checked browser-hosts, and WriteFrankenPHPQuadletDiff mounts ContainerHostsFile without calling ensureFPMHostsFile despite the comment saying it mirrors WriteFPMQuadlet. also client-exec runs without a hosts mount so the shim inherits the host file the same way. happy to open issues for those.

geodro (approved):

PR #1107, opened 2026-07-23
feat(share): serve on a custom Cloudflare-managed domain via --domain

### LINKED ISSUE #1106: share: support a custom domain with Cloudflare Tunnel
`lerd share --cloudflare` only starts a quick tunnel, so every run hands out a fresh random trycloudflare.com URL. That is fine for a one-off peek, but as soon as you want to send a client the same link twice, or point a webhook or OAuth callback at the site, the rotating hostname gets in the way.

Cloudflare already solves this with named tunnels: log in once, create a tunnel, route a hostname you own to it, and the URL is stable for as long as you keep the DNS record. lerd could drive that flow itself so sharing on your own domain is one flag instead of four cloudflared invocations.

Proposal: `lerd share --domain dev.example.com` (implying `--cloudflare`). If cloudflared has no origin cert yet it runs the interactive login, then it creates or reuses a tunnel named after the site, routes the hostname to it, and serves the site through the same host-rewriting proxy the quick tunnel path already uses. The only requirement on the user's side is that the domain's DNS is managed by Cloudflare.

prabowosd:
`lerd share` gets a `--domain` flag that serves the site on a hostname the user owns instead of a random trycloudflare.com URL. Passing `--domain dev.example.com` switches from a quick tunnel to a named one: if cloudflared has never logged in, the interactive browser login runs first, then a tunnel named `lerd-<site>` is created (or reused on later runs), the hostname is routed to it with `cloudflared tunnel route dns`, and the tunnel runs against the same host-rewriting loopback proxy the quick tunnel path already goes through, so Host, Location, and asset URL rewriting all behave identically.

Route DNS is the one non-idempotent step, cloudflared refuses to touch an existing record, so a second run against the same hostname prints a note asking the user to check the CNAME instead of failing the share. Combining `--domain` with a non-Cloudflare tunnel flag is rejected up front, and without the flag the quick tunnel behaviour is byte-for-byte what it was.

Following up on review, `--domain` no longer implies `--cloudflare`. A new `lerd share:tool` command stores a default tunnel tool in the global config (`share.default_tool`), a bare `lerd share` uses it before falling back to auto-detection, and an explicit tool flag still wins per run. `--domain` now only applies when the effective tool is Cloudflare Tunnel, either via `--cloudflare` or the configured default, and errors with a pointer to both otherwise.

Refs #1106

### DIFF
--- a/cmd/lerd/main.go
+++ b/cmd/lerd/main.go
@@ -208,6 +208,7 @@ func main() {
 		root.AddCommand(cmd)
 	}
 	root.AddCommand(cli.NewShareCmd())
+	root.AddCommand(cli.NewShareToolCmd())
 	root.AddCommand(cli.NewDomainCmd())
 	root.AddCommand(cli.NewGroupCmd())
 	root.AddCommand(cli.NewWorkspaceCmd())
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -14,6 +14,11 @@ node:
                           # / node:unmanage; honoured on lerd update so an opt-out
                           # is not undone. Omitted on configs predating it, which
                           # fall back to whatever shims are on disk.
+share:
+  default_tool: cloudflare  # optional. Tunnel tool lerd share uses when no tool
+                            # flag is given: ngrok, cloudflare, expose, serveo or
+                            # localhost-run. Written by lerd share:tool; omitted
+                            # (the default) means auto-detect.
 nginx:
   http_port: 80
   https_port: 443
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -69,7 +69,7 @@ Setup steps include common tasks (composer install, npm install, lerd env) plus 
 | `lerd sites` | Table view of all registered sites |
 | `lerd open [name]` | Open the site in the default browser |
 | `lerd share [name]` | Expose the site publicly via ngrok, cloudflared, or Expose (auto-detected) |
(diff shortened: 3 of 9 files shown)


geodro:
Pushed a commit to this branch, hope that is ok, it sits on top of what you had rather than rewriting any of it.

The main thing is that --domain no longer needs --cloudflare next to it. Custom hostnames only exist for Cloudflare Tunnel, so lerd can work that out by itself, and --domain now also wins over a different default set with share:tool. Pairing it with another tool flag is still an error rather than a silent override.

Three smaller things came out of running it against a real site on my own domain. If a tool pinned with share:tool later loses its binary, a bare lerd share used to fail with just cloudflared not found in PATH, which reads like lerd broke rather than like a stale setting, so the error now says where the choice came from and how to undo it. If the tunnel exists on the account but its credentials json is not on this machine, cloudflared died with tunnel credentials file not found after lerd had already printed a public URL, so the credentials are checked before anything gets announced. And a freshly routed hostname does not resolve straight away, I opened mine too early and my resolver held on to the miss for half an hour while the tunnel was sitting there perfectly healthy, so routing a new record now warns about that.

Also gave share:tool some help, it had no examples and printing the current value did not tell you how to change it, plus a test file since the command shipped without one.

One correction to the description, route dns is actually idempotent when the record already points at the same tunnel, it only refuses when the record aims somewhere else, so the docs say that now. The tolerance you wrote is still right, just a narrower case than either of us described.

geodro (approved):

PR #1148, opened 2026-07-24
feat: service dependencies should accept an installed drop-in

### LINKED ISSUE #1114: Service dependencies should accept an installed drop-in, not just the literal service name
phpMyAdmin's preset declares depends_on mysql, but a user who removed mysql and runs mariadb instead should be able to install and use it too. Today the install preflight happens to accept mariadb, because phpmyadmin co-discovers the mysql and mariadb families through its discover_family directive, but the start paths still start the literal dependency name, so on a mariadb only machine installing phpmyadmin passes preflight and then fails trying to start lerd-mysql. RedisInsight has no discover_family at all, so a valkey only machine is blocked at preflight even though valkey is a drop-in it administers, and pgAdmin with only postgres-pgvector installed hits the same literal start failure.

The store already knows which services are drop-ins for which: mariadb declares env_role mysql, valkey declares env_role redis, and the postgres alternates share family postgres. Dependency satisfaction should be built on that instead of being inferred from discover_family. A depends_on entry names a service, and the dependency is met by that service or by any installed service whose family or env_role names it. The start paths resolve each dependency to whichever satisfier is actually installed and start that one, and when nothing satisfies it the error mentions the accepted alternatives instead of just the literal name.

This keeps depends_on as a plain list of names, so nothing changes in the store schema, older binaries keep parsing the store, and user service YAMLs already on disk keep working without a migration.

raven-pierce:
## Summary

A `depends_on` entry was treated as a literal service name on start, while install preflight only widened satisfaction through the dependent's `discover_family` directives. That let phpMyAdmin install on a MariaDB-only host and then fail trying to start `lerd-mysql`, blocked RedisInsight on a Valkey-only host at preflight, and hit the same start failure for pgAdmin when only a postgres-family alternate like `postgres-pgvector` was installed.

Dependency satisfaction now uses what the store already declares: the named service, or any installed service whose `family` or `env_role` names it. Start resolves to that satisfier and brings it up. When nothing satisfies a dependency, the error lists the accepted alternatives. `depends_on` stays a plain list of names, so older binaries and on-disk user YAMLs need no migration, and `discover_family` remains admin host discovery only.

The web UI, CLI, and TUI also show the resolved satisfier's preset name (mariadb, not mariadb-11-8) and light the status from that unit, so a drop-in no longer looks missing next to a gray dot.

Closes #1114.


geodro (changes requested):
The consolidation is the right call, having CLI, UI, TUI and MCP all go through one serviceops lifecycle is worth the churn on its own. I ran the gate and then put the branch on the Ubuntu and Fedora VMs, and the headline case really does work now: with no mysql installed at all, phpMyAdmin passes preflight, starts, renders PMA_HOSTS as lerd-mariadb-11-8 and serves a real logged in page. Stopping mysql with mariadb still up correctly leaves phpMyAdmin alive and rewrites its host list, and stopping the last satisfier cascades properly. Good work. Four things I want changed before it goes in though, two of which I reproduced on the VMs.

The one that will bite a user is autoStopUnusedServices, which now calls StopService, and StopService sets the paused flag, which means "the user manually stopped this" so lerd start skips it. I ran this side by side on the same VM with only the binary changing. On main, pausing every site auto stops redis and mailpit and paused-services.yaml is untouched, and a later lerd start brings both back. On this branch the same sequence writes redis and mailpit into paused-services.yaml, unpausing the site restarts them but leaves the flag set, and after a stop and start neither one comes up, the containers do not even exist, while the site .env still points at lerd-redis and lerd-mailpit. Worth keeping the old split, either a variant of StopService that skips the flag or just have autoStopUnusedServices call StopWithDependents like it used to.

RedisInsight and mongo express are the awkward pair. Neither declares discover_family, they pin their connection host in the environment block, RI_REDIS_HOST is lerd-redis and mongo express hardcodes lerd-mongo in the URL. I removed redis on the Fedora box, installed valkey, and redisinsight now installs and reports active, but RI_REDIS_HOST is still lerd-redis, getent for it inside the container exits 2, and the container log says the connection timed out. So the case in the description is unblocked but lands broken. That is what TestMissingPresetDependencies_NonDiscoveringDepStaysStrict was guarding, its comment spelled out why, and it got deleted and replaced with a test asserting the opposite. I think this needs a services store change alongside it so those two resolve their host to whatever satisfied the dependency, otherwise the strictness should stay for deps the consumer cannot discover.

Third, the accepted alternatives never show up where it matters. On a clean box the error is still the bare "requires service(s) mysql to be installed first". dropInPresets scans ListPresets, which is the embedded bundle plus the local store cache, and the embedded bundle is only mailpit, meilisearch, mysql, postgres, redis and rustfs, so mariadb and valkey are invisible until the user has already fetched them, which is exactly when the hint stops being useful. Both docs pages and the description promise it lists them.

Fourth, stop failures now report success everywhere. StopService always returns nil, StopWithDependents discards the StopUnit error and does nothing at all when the unit reads as inactive, and both the UI handler and MCP take that nil as OK. The old UI code called StopUnit unconditionally with a comment saying it was there to cover exactly the status lag case, and this PR adds ContainerRunning fallbacks for lag in RegenerateFamilyConsumers, so let's be consistent and return a real error here.

Smaller one on the display side. DependencyDisplayName prefers the literal name whenever it is installed, running or not. With mysql installed but stopped and mariadb actually backing phpMyAdmin, service list still prints "depends on: mysql" and the API returns mysql while reporting it inactive, so the dashboard shows phpMyAdmin hanging off a grey dot while the thing serving it is up. That is the symptom the third paragraph of the description set out to remove, just reached from the other side, so resolution for display probably wants to prefer a running satisfier.

Two notes rather than asks. DependencyDisplayNames runs per service in the services list response with each dep resolving twice, each resolution rereading every custom service yaml and loading every default preset, while uniqueFamilyHosts forks podman inspect per family member per directive, three times over for phpMyAdmin, and the dashboard polls that endpoint. And config.ServiceRunning as a package global wired from a serviceops init means importing serviceops changes how config renders quadlets, the TestMain that unsets it is the tell, a parameter on ResolveDynamicEnv would be less surprising.

For what it's worth I also checked whether the running only host filter causes phpMyAdmin to be started wrong and then bounced on every lerd start, and it does not, ReconcileServices only regenerates missing or drifted quadlets so the empty list never gets written before the bulk start.

geodro:
I swapped out the resolve_dep approach for the redisinsight and mongo express host. The trouble with putting it in a dynamic_env directive is that any binary that does not know the directive hard errors on it at quadlet generation, so the moment those presets landed in the store every install on 1.30.1 or older would break trying to bring them up, which is exactly the backward compatibility 1114 was trying to keep.

So I moved the resolution into the binary and drove it off depends_on. It rewrites a pinned lerd-<dep> host to whatever actually satisfies the dependency when it generates the quadlet, so redisinsight just keeps its plain RI_REDIS_HOST of lerd-redis in the store and ends up on lerd-valkey when valkey is the thing satisfying redis. Nothing special in the yaml for an old binary to trip over, and it behaves the same as before, one satisfier, preferring one that is running.

I pushed the matching change to the services PR too, the two presets go back to plain environment and I kept only the env_role additions in the index, since those are just metadata an old binary ignores.

geodro (approved):

PR #1176, opened 2026-07-25
feat(services): add expand_family dynamic_env directive

### LINKED ISSUE #1147: dynamic_env needs a per-member expansion so RedisInsight can preconfigure every redis family connection
The RedisInsight preset preconfigures a single hardcoded connection, RI_REDIS_HOST pointing at lerd-redis. An install running valkey instead of redis, or both side by side, opens RedisInsight with those containers missing even though the preset administers them. RedisInsight can preconfigure any number of databases through numbered env var sets, RI_REDIS_HOST with a connection id suffix plus matching port and alias vars per connection.

The dynamic_env engine cannot express that today. discover_family and repeat_family both emit one comma joined list into a single key, which fits phpMyAdmin's PMA_HOSTS but not one var set per connection. It needs a per member variant: a directive that takes the same family list and writes one env var set per installed member with the key numbered, so the redisinsight preset can declare its RI_REDIS_* vars declaratively and every installed redis and valkey shows up wired on first open. The preset change itself is a lerd-env/services publish once the directive ships.

prabowosd:
discover_family and repeat_family both collapse a family into a single comma joined value, which is exactly what phpMyAdmin's PMA_HOSTS wants and exactly wrong for a tool that reads numbered env var sets. RedisInsight is the case in point: it preconfigures databases through RI_REDIS_HOST plus a connection id suffix with matching port and alias vars, so its preset had to hardcode one connection at lerd-redis, and an install running Valkey instead, or both engines side by side, opened the UI with those containers missing even though the preset administers them.

expand_family:<families>=<template> writes one env var per running member, the declared key suffixed _1 through _N. {host} expands to the container hostname and {name} to the service name, and a template with no placeholder is repeated per member, which covers the constant port. The unsuffixed key is never written. Members come from the same sorted, deduplicated, running only host list the other two directives already use, so the numbering is stable as long as the set of running members is.

Family consumer detection now reads the family list off any of the three directives instead of discover_family alone. Without that, a preset wired only with expand_family would never count as a consumer, so its quadlet would not be regenerated and bounced when a redis or valkey is installed, removed, started or stopped, and the drop in dependency check would refuse Valkey for a redis dependency.

The redisinsight preset that uses this is a lerd-env/services publish, separate from here.

Closes #1147


geodro (changes requested):
The directive itself works, I ran it on three installs against redis and valkey side by side and against a valkey only setup, the numbered var sets come out right, RedisInsight preconfigures every member and the regen fires on install, remove, start and stop, including for a preset wired only with expand_family. The consumer detection fix for repeat_family is a nice catch too.

What stops me approving is the rollout story from my comment on #1147. A binary that predates this fails hard on the unknown directive, I put the expand_family yaml on a 1.30.1 install and redisinsight just refuses to start with "unknown dynamic_env directive". And because a preset backed service re syncs its yaml from the store definition on every start, the moment the redisinsight preset publishes, every older install hits that error on the next restart without ever asking for an update. That comment also floated the json file route, RI_PRE_SETUP_DATABASES_PATH, as the alternative that degrades instead of breaking on old binaries, and I'd still like a sentence on why the directive wins before we commit to it, even if the answer is just that a generic directive serves every future preset and the json file only serves this one.

Assuming the directive is the way to go, can you add the tolerance here as well, have ResolveDynamicEnv warn and skip an unknown directive instead of erroring, so the quadlet still generates with whatever static environment the preset carries. That turns the failure mode from a dead service into a degraded one, which is the same property that made the depends_on rewrite in 1114 safe to ship. The preset publish then waits until a tolerant binary is the deployed floor, but at least this PR stops adding to the problem it needs to wait for.

geodro:
The redisinsight preset publish this enables is tracked in lerd-env/services#12.

prabowosd:
Tolerance added in 17bfd0a3. `ResolveDynamicEnv` now warns and skips a directive it does not recognise instead of returning an error, so the quadlet still renders with whatever static `environment` the preset carries and the service starts degraded rather than not at all. Malformed payloads on a directive the binary does know, `expand_family:redis` with no `=`, still error, since that is an author mistake an older binary cannot be shielded from anyway. Both docs pages say so, and `TestResolveDynamicEnv_UnknownDirective` now asserts the skip, the surviving static env and the warning text.

Worth noting this makes `dynamic_env` consistent with the rest of the preset surface rather than inventing a new rule: `PresetFiles` already skips a mount naming an unknown `generator` for exactly this reason, and `RewriteDependencyHosts` was deliberately built on `depends_on` so the published YAML carries nothing an old binary rejects. `dynamic_env` was the last place where a forward-looking store definition could hard-fail.

On directive versus `RI_PRE_SETUP_DATABASES_PATH`: the json file does not actually avoid the version floor. A preset cannot ship its own generator, only reference a built-in one, so the file route needs a new built-in generator and therefore a lerd release too. The difference between the two was only the failure mode on an old binary, and that difference is what this commit removes. What is left is that the directive is generic where the generator is not: any future preset reading numbered env var sets is served by YAML alone, whereas a `pre_setup_databases` generator would bake RedisInsight's json schema into the binary and serve that one tool. Env vars are also RedisInsight's documented preconfiguration contract, so there is no file to keep in sync when they change it.

The preset publish still waits for a tolerant binary to be the deployed floor, but nothing added from here on compounds the problem.

PR #1180, opened 2026-07-25
feat(nplusone): always name where the duplicate queries ran

### LINKED ISSUE #1171: N+1 notifications should always say where the queries happened
The N+1 push notification names the route only when the query events carry an HTTP request context, so a web hit reads "GET /orders ran a similar query 3×" but everything else falls back to the generic "Ran a similar query 3× in one request". In practice that covers all CLI invocations, artisan commands, tinker sessions and scheduled tasks, since the debug bridge only stamps the request field for FPM and the worker name is only present for opted-in worker capture. The user gets a warning with no way to tell where to start looking.

Every notification should point somewhere. The bridge can stamp a command context for CLI invocations (the script name and arguments), and the notification can use that the same way it uses the worker command today. As a last resort the query event already carries the originating file and line in its source field, so even an event with no request and no command can fall back to "app/Jobs/SyncUsers.php:42 ran a similar query 3×" instead of the generic line.

prabowosd:
The N+1 notification named a route only when the query events carried an HTTP request context, so a web hit read "GET /orders ran a similar query 3×" while everything else fell back to "Ran a similar query 3× in one request". Since the bridge stamps request only for FPM and worker only for opted-in worker capture, that generic line covered every CLI invocation, artisan command, tinker session and scheduled task, which is a warning with nowhere to start looking.

The bridge now stamps a command context on every cli-SAPI process, the script basename plus its arguments ("artisan queue:work --queue=high"), capped at 120 characters so a long argument list cannot bloat the context of every event a process emits. It is the CLI counterpart of the request field, and the notification consumes it the same way it already consumes the worker command.

The body picks the most specific label available: worker, then command, then request, and finally the file:line the query event already carries, so even an event with neither a request nor a command reads "app/Jobs/SyncUsers.php:42 ran a similar query 3×" instead of the generic line. The dedup key gained the same command arm, because without it every console invocation shares one key and only the first artisan command of a session could ever warn.

Closes #1171


geodro:
Good idea and the Go side was solid, but on a real install the command never actually showed up. The collector's context() only feeds events the collector itself emits, which is mostly dumps. Laravel query events come from the Laravel adapter, which carries its own copy of context(), and other frameworks get their queries emitted straight from the devtools extension, so the N+1 warning never saw a command and console runs still shared one dedup key. I pushed a commit that stamps the same command in the adapter, plus a harness test that runs the real adapter under a host php the way the collector test does.

Driving it on a real site surfaced a second thing, tinker put its whole inline snippet into the notification body, since the code is just another argv argument. The second commit condenses the command, any argument value longer than 32 characters is elided down to its flag, so queue:work keeps its queue name but a tinker run reads artisan tinker --execute=... The file and line fallback is gone too, the label now names the run, worker, command or route, and the exact query origin already rides the event in the Debug lens. One consequence, all tinker one liners now share a dedup bucket, one warning per session, which matches the once per route behaviour on the web side.

The remaining gap is the extension path, queries from frameworks without an adapter build their ctx in the extension and still carry no command, so they fall back to the generic line. That fix ships in the base images and needs a rebuild, it will get its own issue.

geodro (approved):

PR #1235, opened 2026-07-30
Fix client-shim loopback host routing and anchor worker proxy paths

### LINKED ISSUE #1233: Client-tool shims fail to connect when pointed at the service's own published host port
Every client-tool shim lerd installs (psql, pg_dump, mysql, mysqldump) execs the underlying tool inside a throwaway container joined only to the lerd network. A hostless call like psql -l correctly routes to the internal lerd-<service> container name and works. But an explicit -h 127.0.0.1 or -h localhost at the port lerd itself published, the normal way to reach it from a script, an IDE, or muscle memory, fails with connection refused every time, because inside that throwaway container the loopback address is its own, not the host's. podman ps shows the service running and a hostless client call can list its databases just fine, which makes the failure look intermittent or environment specific when it's actually a deterministic gap in how the shim decides what counts as an external host.

### LINKED ISSUE #1234: Worker proxy nginx location block matches unrelated routes sharing the same path prefix
The nginx vhost templates render a worker's proxy path as a bare location /app block. nginx treats a location with no modifier as a prefix match, so it also catches any route that merely starts with the same characters, like /appstore or /apples, and sends them to the worker's proxy backend instead of letting the site serve them normally. Framework store data can also legitimately contain regex metacharacters in that path, /socket.io being a common real one, which matters once the fix anchors the match as a regex, since an unescaped dot would then match any character instead of a literal one.

retr0ripper:
This bundles two unrelated fixes from the same investigation session, kept in one branch since they're both small.

The first three commits fix the client-tool shim bug: psql, pg_dump, mysql, and mysqldump all fail with connection refused when given an explicit -h 127.0.0.1 or -h localhost at a port lerd itself published, because the shim always execs the tool inside a throwaway container on its own network, where that loopback address doesn't reach the host. resolveLoopbackTarget now recognises when the given loopback host and port match one of lerd's own installed services and routes it the same way a hostless call already does, using two small new helpers, HostPortsFor to read a service's actual published ports and ToolCandidates to tell same-family instances apart when more than one is installed. A loopback host matching nothing lerd owns, or any other host, still passes straight through unchanged.

The last commit fixes the worker proxy location block in the nginx vhost templates, which used a bare location /app that nginx treats as a prefix match, so it also caught unrelated routes sharing the same prefix and sent them to the proxy backend instead of the site. Anchoring it as a regex fixes that, and since the path now reaches nginx as a regex, it's escaped first so a literal path like /socket.io isn't misread as a wildcard.

Refs #1233, Refs #1234


geodro:
Nice one, this pair was fiddly to track down and the loopback matching is the shape I would have picked myself, only the services that actually expose the tool, so a mysql port can never wander into postgres.

One thing before I merge. The path lands as ^path(/|$), so a definition that writes /app/ with the slash on the end renders ^/app/(/|$) and matches nothing at all, not the bare path, not anything under it. Nothing in the store does that today so nothing is broken, but your new docs invite people to write the path however reads naturally, and the first one with a trailing slash gets a proxy that quietly never fires. Trimming it before the escape is all it takes.

Also worth knowing, going to location ~ makes it a regex location, and nginx runs those in file order ahead of the prefix it remembered, so the proxy now outranks the php block and the dotfile deny under that path. For reverb on /app that is fine, probably what you want anyway, I just wanted it said out loud since anchoring was the goal and this rode along with it.

Ping me when the slash is trimmed and I will get it in. Thanks for the two clear issues as well, they made this an easy read.

retr0ripper:
Hi @geodro,
Done. I've trimmed the trailing slash before escaping, and added a note to the docs explicitly calling out the regex location precedence over the PHP and dotfile blocks.

Ready for another look!

geodro:
Slash trim is right, `/app/` comes out as `^/app(/|$)` now, and putting the precedence note in the docs was the better call, that belongs there rather than buried in this thread.

I said ping me and I will get it in, and then went through the shim half properly and found something, so sorry for moving the goalposts this late.

The host decision in runClientExec used to run through argsSpecifyHost, which caught three shapes, the flag forms, a connection URI through isConnURI, and a libpq conninfo string through looksLikeConninfo. hostFlagSpan only covers the flag forms, which its own comment says plainly enough, so the other two now read as no host at all. argsSpecifyHost has no callers left in production, only its own test, which is why the suite stayed green over it.

What that means in practice is that a URI or a conninfo string falls into the local default branch, so we prepend `-h lerd-<service>` and set localCredsEnv on a connection that was never ours to touch.

I put your branch on a clean Ubuntu 26.04 box to be sure rather than reason about it, installed it the normal way, started postgres, added the psql shim with lerd shims add, then stood up a second postgres with its own user and password to play the part of an external server. Same command through the real shim, twice, the only difference being the binary.

On 1.31.0 it prompts, Password for user extuser, then gives up with fe_sendauth: no password supplied. Nothing of ours went anywhere, which is the behaviour we want.

On your branch the prompt never appears, and it comes back with FATAL: password authentication failed for user extuser. No prompt means a password was supplied for us, and the external server's own log has exactly one authentication failure in it, from that run and not the other. So PGPASSWORD=lerd genuinely left the machine and arrived at a host lerd does not own. It is a known constant so the blast radius is small, but that is the part I do not want to ship.

Tidiest fix I think is to keep the URI and conninfo checks inside resolveLoopbackTarget and return hostGiven true for them before it looks at a single flag, so one function owns the whole answer and argsSpecifyHost can go for good.

One more while you are in there, same species as the trailing slash so it may as well ride along. A path of just `/` trims to empty, gets set back to `/`, and renders `^/(/|$)`, which matches `/` and `//` and nothing else, so a worker mounted at the root would proxy nothing underneath it. Nothing in the store writes that today, same as before, but the docs now invite people to write the path however reads naturally and that is the other way to trip it.

Branch is behind main too, worth a rebase before I take it.

PR #1237, opened 2026-07-30
Repair mkcert CA trust on macOS instead of trusting presence alone

### LINKED ISSUE #1236: mkcert CA reported trusted on macOS by presence alone, not actual trust settings
CATrusted() only checked whether mkcert's root CA was present in the keychain, via security find-certificate, never whether it actually carried trust settings. A certificate can sit in the keychain with its trust settings cleared independent of the certificate item itself, most commonly after a macOS update, and find-certificate reports that identically to a genuinely trusted CA. lerd would then treat a reinstall as a no-op and skip the sudo announcement, leaving the CA silently untrusted and every site's HTTPS certificate invalid in the browser, even though podman and the vhost itself looked completely healthy. mkcert's own "already installed" self-check has the same blind spot, since it verifies the self-signed root against itself, which succeeds regardless of actual trust settings, so it can't be relied on to notice or repair the drift on its own.

retr0ripper:
platformTrustCheck now parses security trust-settings-export -d, which security itself keys by each certificate's SHA-1 fingerprint and which only carries a trustSettings array for a certificate actually configured as trusted, the distinction find-certificate can't make. The old presence-only check is kept under platformPresenceCheck for the one thing it's still useful for, telling CAPresentButUntrusted apart from a CA that was never installed at all.

Since mkcert's own self-check has the same blind spot, ensureMkcertCA now repairs trust directly with the same security add-trusted-cert command mkcert itself uses, before mkcert gets a chance to wrongly decide there's nothing to do.

Refs #1236


geodro:
Hey @retr0ripper some CI issues

retr0ripper:
Hi @geodro ,
Dug into the CI failure, turns out it had nothing to do with the mkcert changes. The Linux job was failing on TempDir cleanup in internal/podman and internal/xdebugops, both tripping over the same container storage overlay layer. DetectHostGatewayIP has a last resort fallback that runs a real podman run of the alpine image when every cheaper probe comes up empty, which is always true in these tests since lerd nginx never actually runs. Both test files tried to block real podman by emptying PATH, but PodmanBin also tries a couple of hardcoded absolute paths as a fallback, and those skip PATH lookup entirely, so whichever of those exists on the machine running the test let a real podman through and left behind overlay layers that a plain RemoveAll can't clean up. Fixed the podman package test to stub its own exec seam directly, and the xdebugops one to pre seed the hosts file so it never reaches that fallback at all. Pushed the fix and reopened this.

PR #1267, opened 2026-08-03
fix(mcp): declare items schema on array-typed tool parameters

### LINKED ISSUE #1265: [VSCode] Lerd MCP Error: tool parameters array type must have items
After engaging a Chat conversation in VSCode, it throws: `Failed to validate tool mcp_lerd_diag: Error: tool parameters array type must have items.` pretty much immediately. Not sure if VSCode is just not supported yet or a configuration issue.

Config:
```json
"servers": {
  "lerd": {
    "args": [
      "mcp"
    ],
    "command": "lerd",
    "type": "stdio"
  },
  // ...
}
```

MCP Lerd Output:
```log
2026-08-03 11:21:30.032 [info] Starting server lerd
2026-08-03 11:21:30.032 [info] Connection state: Starting
2026-08-03 11:21:30.032 [info] Starting server from LocalProcess extension host
2026-08-03 11:21:30.045 [info] Connection state: Starting
2026-08-03 11:21:30.047 [info] Connection state: Running
2026-08-03 11:21:30.250 [info] Discovered 12 tools
// nothing follows...
```

VSCode Information:
```yaml
Version: 1.131.0
Commit: e4c7e7b1d6d060162f4aa7f8225271b67ce1df75
Date: 2026-07-28T10:51:25Z
Electron: 42.7.0
ElectronBuildId: 14710161
Chromium: 148.0.7778.280
Node.js: 24.18.0
V8: 14.8.178.38-electron.0
OS: Linux x64 7.1.5-201.fc44.x86_64
```

JoeJoeflyn:
Every MCP tool property with type: array was emitted without an items field,
which the MCP spec requires to describe the element type. VSCode strict-validates
this and rejects every tool with "tool parameters array type must have items",
while lenient clients (Claude, Cursor) ignored it.

The mcpProp struct now carries an Items field, and every array-typed property
sets it to stringItems (all lerd MCP arrays are string arrays). The tool list
size ceiling rises from 22450 to 22900 for the added schema content.

Closes #1265


geodro:
pushed a small follow up onto your branch, hope that is alright. the items field was exactly the right call, that is what vscode is rejecting. two of the arrays are not string arrays though, framework setup and logs are both parsed as objects by the handler, setup as label, command, default and an optional check, logs as path plus an optional format. with string items a strict client sends strings and the map assertion drops every element silently, so the call says the framework was saved and the yaml comes back with no setup and no logs. added an objectItems var for those two and extended your test to check the element type rather than just that items exists. everything else was spot on, thanks for tracking this down

JoeJoeflyn:
good call, thanks for the fix

PR #1269, opened 2026-08-03
fix(node): preserve npm_config_prefix in global wrappers after activation

### LINKED ISSUE #1266: Codex global shim should preserve Lerd npm prefix
npm global wrappers activate Node through `fnm exec`, which prepends fnm’s npm binary and prefix. Tools such as `codex doctor` then report that updates would target a different installation even though Codex exists only in Lerd’s `node-global` prefix.

Generated wrappers should pass Lerd’s `npm_config_prefix` after Node-manager activation. This must work for both fnm and nvm without exposing the prefix before `nvm use`.

Reproduction: install Codex through `lerd npm install -g @openai/codex`, then run `codex doctor --json`. The installation and update checks fail. Running with `NPM_CONFIG_PREFIX=~/.local/share/lerd/node-global` makes every doctor check pass.

JoeJoeflyn:
npm global wrappers activate Node through fnm or nvm, which clobbers npm_config_prefix. Tools like codex doctor then see the wrong installation. Added ExecPrefixWithEnv to the Manager interface so wrappers inject the prefix after activation, fnm via `env` after `--`, nvm via `export` after `nvm use`.

Closes #1266


geodro:
pushed one more commit onto your branch. the approach here is right, injecting after activation is exactly the fix, and putting env after the -- for fnm and the export after nvm use is the correct placement.

one thing though, the nvm path quotes its values because it goes through nvmExports, but the fnm path joined them raw. the prefix gets interpolated straight into a /bin/sh wrapper, so if the data dir has a space in it the path splits in two and env takes the tail as the command. tested it on an ubuntu guest with the data dir moved somewhere with a space, and the wrapper dies with env: 'dir/lerd/node-global': No such file or directory, so the real binary never runs at all, which is worse than the missing prefix. quoting it the same way nvm does fixes it and the binary then gets the prefix intact.

had to move your InjectsAfterActivation assertion to the quoted form since the output changed, and added one that runs the fragment through a real sh rather than matching a substring, since a substring match is what let this slip past in the first place. rest of it was solid, thanks

PR #1302, opened 2026-08-05
fix(dns): detect missing dnsmasq binary on NetworkManager-dnsmasq hosts

raihanadf:
## Summary
- NetworkManager's `dns=dnsmasq` plugin needs a local `dnsmasq` binary to work. On Arch/its derivatives nothing pulls it in by default, so lerd configured NetworkManager "successfully" while `.test` silently kept resolving through the raw upstream DNS (only visible via `journalctl -u NetworkManager`).
- `setupNetworkManager` now checks for the binary and fails with an install hint instead of restarting NetworkManager into a broken state.
- `lerd doctor`'s "resolver hookup" check now catches this too, instead of falsely reporting OK.

## Test plan
- [x] `go build`, `go vet`, `go test ./internal/dns/...` all pass
- [x] Reproduced and fixed on a live Arch host


geodro:
Good catch, this closed a real silent failure. I reproduced it on an Arch VM, without dnsmasq the old code restarts NetworkManager into a config it can't run and the only trace is one journal warning, with your guard it stops cleanly and doctor points at the exact fix. I pushed one commit on top of your branch. LookPath only searches the caller's PATH, and on Debian apt puts dnsmasq in /usr/sbin which a normal user's PATH doesn't include, while NetworkManager finds the binary through its compile time path just fine. So on a correctly set up Debian host the guard would have refused to configure NetworkManager and doctor would have reported a failure on a working install. The lookup now falls back to statting the usual sbin locations when PATH misses, and it lives in one shared helper used by both the setup guard and the doctor rung so the two can never disagree. Verified the whole thing across Arch, Fedora, Silverblue and Ubuntu guests, your scenario end to end on the Arch one, no behavior change anywhere systemd-resolved is in charge. Thanks for digging this out of journalctl.

PR #1427, opened 2026-08-08
feat(watcher): tear the environment down when the OS signals a logout

### LINKED ISSUE #1407: feat(cli): keep lerd daemon alive so macOS can gracefully stop Podman
lerd starts the Podman Machine and containers at macOS login through lerd-autostart, but the lerd start process exits as soon as startup finishes, leaving the VM running without a process that launchd can signal when macOS shuts down. The result is that the Podman Machine can be killed abruptly, and lerd-postgres-timescaledb sees an unclean PostgreSQL shutdown and spends several minutes in crash recovery and fsyncing data on the next boot.

It should run a long-lived lerd daemon instead. The daemon starts the same environment as lerd start, then stays alive waiting for SIGTERM or SIGINT. When the service manager asks it to stop, it runs the normal lerd stop flow and waits for that shutdown to finish before exiting, giving the containers and Podman Machine a chance to shut down cleanly. The daemon should reuse the existing start and stop implementation rather than spawning lerd start and lerd stop as child commands, so there is one lifecycle implementation and no recursive service behavior.

The autostart service should invoke lerd daemon instead of lerd start, with Restart=on-failure so an unexpected daemon failure is still recovered by the service manager. The command itself is an internal implementation detail, so it should be registered as hidden and not become another user-facing command in the normal lerd --help output.

The shutdown path needs to wait for lerd stop to complete before the daemon exits. Receiving SIGTERM, starting the stop operation and immediately exiting would leave the VM exposed to the same abrupt termination this change is intended to fix. The handler should also tolerate an already-stopped or partially stopped environment and avoid running shutdown concurrently if multiple signals arrive.

The existing lerd start and lerd stop commands keep their current behavior. Only the macOS autostart lifecycle changes from a fire-and-forget start process to a supervised daemon, so manually starting and stopping Lerd does not acquire a persistent background process.

The main files are internal/cli/daemon.go for the new hidden command, the command registration in cmd/lerd/main.go (or the existing command registry), and internal/systemd/units/lerd-autostart.service for the service change from lerd start to lerd daemon.

Verification should include running the daemon directly and confirming it remains alive after startup, sending it SIGTERM and checking that the normal stop sequence completes before it exits, then rebooting macOS through the actual autostart service. After the reboot, lerd-postgres-timescaledb should come up without the multi-minute PostgreSQL crash-recovery/fsync phase. Repeated reboot/shutdown cycles should produce the same clean result.

### LINKED ISSUE #1428: fix(ui): dashboard Quit leaves the Podman Machine running
The Quit Lerd button in the dashboard calls cli.RunQuit inside the lerd-ui process itself. The teardown stops containers and workers first, then walks the host process units in order, and that list starts with lerd-ui. Stopping lerd-ui asks the service manager to bootout the job the calling process is running in, which delivers SIGTERM, and lerd-ui has a handler that stops its tunnels and exits immediately. The process dies on the first unit of the loop.

Everything after that never runs. lerd-watcher, lerd-tray and lerd-dns are left up, the Podman Machine VM is never stopped, and the stray tray process is never killed. Quitting from the dashboard therefore leaves the VM running, which is the same abrupt VM termination that leaves lerd-postgres-timescaledb in a multi-minute PostgreSQL crash recovery on the next boot. Linux behaves the same way, since stopping lerd-ui there delivers the same signal to the same handler.

The tray does not have this problem because its Quit menu item shells out to lerd quit as a separate process, so the teardown outlives the units it stops.

The dashboard should do the same and spawn the quit as a detached child process rather than run it inside the process it has to kill. The alternative is to keep it in process and pass lerd-ui to the skip list lifecycle.Quit accepts once #1427 lands, then have lerd-ui exit after the rest of the teardown finishes, but that leaves the UI job loaded and depends on the exit not being treated as a failure worth respawning.

The main files are internal/ui/server.go for the handler and internal/ui/shutdown.go for the signal handler that ends the process, with internal/tray/tray.go as the working reference.

Verification should include quitting from the dashboard on macOS and confirming that lerd-watcher, lerd-tray and lerd-dns are all stopped afterwards and that podman machine list reports the VM stopped, then confirming lerd-postgres-timescaledb starts without a crash recovery phase on the next lerd start. The same check on Linux should confirm the remaining units stop.

retr0ripper:
The Podman Machine VM was being killed abruptly when macOS restarts, so lerd-postgres-timescaledb came back into a multi-minute PostgreSQL crash recovery replaying its write-ahead log. Nothing stopped the VM first because no lerd process was acting on the SIGTERM launchd sends at logout.

The issue proposed a new hidden lerd daemon command wired into lerd-autostart. That turned out not to be necessary. The watcher already runs under launchd for the whole session with RunAtLoad, so it is already the long lived process that receives the signal, and hooking it avoids a second daemon and a new user facing command. It also covers the case where autostart is disabled and lerd was started by hand, which an autostart only daemon would miss.

The watcher now runs the same teardown lerd quit does before it exits: containers and workers stop, then the Podman Machine, then the remaining host processes. The VM goes before the host processes because it is the only step whose loss costs anything, and lerd-ui, lerd-tray and lerd-dns are being terminated by launchd anyway.

The teardown logic moves out of internal/cli into internal/lifecycle so the watcher and the CLI share one implementation rather than the daemon shelling out to lerd stop.

Two things make the signal ambiguous and both are handled. The watcher must never stop its own unit, because asking launchd to bootout the job it is running inside blocks until that process exits, which it cannot do from inside the call, so the teardown would hang until the grace expired and never reach the VM. And lerd install, lerd update and lerd quit all stop the watcher too, arriving as exactly the same SIGTERM a logout does, so every path that stops the watcher marks it first and the watcher exits without tearing anything down when it sees that marker.

launchd gives these jobs five seconds before SIGKILL, which a full teardown does not fit in, so the watcher plist alone gets a 180 second ExitTimeOut and every other job keeps the default. It needs that much because the containers stop before the VM does and a database is entitled to the stop_timeout its service declares, so a grace sized for the containers alone would already be spent by the time the VM stop began, and launchd would kill the watcher partway through the one step this whole sequence exists to protect.

The teardown is macOS only. Linux has nothing for it to protect, because podman runs natively so no VM outlives the session, and systemd already stops the container units on shutdown honouring the stop timeout each one declares. A reboot on Ubuntu 26.04 with the teardown disabled leaves a redis queue that existed only in memory fully intact, because redis receives a clean SIGTERM and saves its final snapshot before exiting, and mysql comes back with no dirty recovery. Running it there would only cost something, since every SIGTERM the watcher receives looks alike and only lerd's own stops carry a marker, so a plain systemctl --user stop lerd-watcher would stop every container, lerd-ui and lerd-dns with it.

Two smaller things ride along because the same shutdown path exposed them. The plists for the jobs that can bring the VM up, meaning lerd-autostart, lerd-ui, lerd-tray and the watcher, now set AbandonProcessGroup, because vfkit and gvproxy reparent to init but keep the process group of whatever started them, and launchd kills whatever remains in a job's group once that job exits. And the macOS exec worker self heal now stands down while lerd is intentionally stopped, since a tick landing inside the teardown would read every dead PID as drift and start the workers back up against it.

Closes #1407


geodro:
Really nice piece of work, the marker funnel in particular is the right shape.

Pushed four small fixes on top. The one that mattered most is the exec worker self heal, it ticks every minute inside the watcher, so a tick landing mid teardown sees every worker's dead PID as drift and starts them all back up while the machine is going down. It stands down now while the stop marker is set, same as the reload watcher already did.

SIGINT was being read as a logout too, so Ctrl-C on a hand run lerd watch took the whole stack with it, only SIGTERM counts now. There was also one way into kickstart -k with no bootout in front of it, which kills the job first and so read as a logout during a plain start, that is marked now. And Quit got a small hook so the tray cleanup lands before podman machine stop rather than after it.

Tested end to end on Ubuntu 26.04 against a real Laravel site, an unmarked stop tears everything down in 6s, lerd install leaves it all running, SIGINT exits clean. Could you run the macOS half when you get a chance? That is the part I cannot reach from here, especially whether launchd really honours the 60s ExitTimeOut during a shutdown rather than loginwindow cutting it short.

One thing worth deciding before merge, and it is your call. On Linux the handler cannot fire at logout, lerd enables systemd linger so the user manager survives the session, and at reboot systemd is already stopping every lerd unit itself. What it does add is that systemctl --user restart lerd-watcher now tears the whole stack down, I hit that by accident cleaning up the VM. Gating the signal handler to darwin would drop that with nothing lost, happy to push it if you agree.

geodro (dismissed):

geodro (dismissed):

geodro (dismissed):

PR #1466, opened 2026-08-11
feat(UI): add copy path button to trace blocks

raisilhamn:
First of all, thank you for creating such an amazing tool! I’d really appreciate it if this PR could be merged, as it makes it much easier to copy the file path from debug queries.

The commit adds a copy button next to the primary file path in the query trace. It copies the `file:line` reference to the clipboard and provides a brief checkmark confirmation after copying.

Thank you .


geodro (approved):

geodro:
@raisilhamn thank you!

PR #1485, opened 2026-08-13
feat(doctor): sweep the linked sites, fold check into site:doctor

### LINKED ISSUE #1358: feat(doctor): one command that answers whether anything is wrong, sites included
There are three health commands and none of them knows about the others. lerd check validates .lerd.yaml, services and the PHP version for the project in the current directory. lerd site:doctor runs the app level checks for one site. lerd doctor covers the environment, podman, systemd, DNS, ports, images, config, and reports how much podman disk is reclaimable without offering to reclaim it. A user who wants to know whether anything is wrong has to run all three, and the one people reach for first, doctor, is the one that never looks at their sites.

lerd doctor should finish by sweeping the linked sites through the site level checks, summarising per site and pointing at site:doctor <domain> for the detail, so the broad command is actually broad. lerd check is a subset of what site:doctor already validates, so it folds in there and stops being a third entry point. And since doctor already computed the reclaimable disk, --fix should offer the cleanup rather than describing it.

The sweep has to stay quick and not turn doctor into a minute long command, so the site pass wants to be the cheap checks by default with the expensive ones, composer audit and the timing lookups, left to site:doctor on a named site.

prabowosd:
There are three health commands and none of them knows about the others, so anyone who wants to know whether something is wrong has to run all three, and the one people reach for first never looks at their sites. `lerd doctor` now ends with a `[Sites]` section: every linked site runs through the site-level checks and gets one line, the counts plus the first thing that is wrong, with the command to run for the detail. A machine where podman, DNS and the ports are all fine while three sites are failing no longer reports itself healthy.

The sweep has to stay quick, so it runs the cheap half of the site doctor and leaves the framework command checks, `composer audit`, `npm audit` and the response-time lookup to `site:doctor` on a named site. The one expensive check it keeps is whether the site's database exists, because that is what catches a site 500ing on every request, and it is no longer expensive: one engine's database list is reused for a few seconds rather than costing a container exec per site. On a fourteen site machine the sweep adds about two seconds to a doctor run.

`lerd check` was the third entry point and everything it validated is about the project rather than the environment, so `.lerd.yaml` validation moves into the site doctor as its own check. Every surface that runs the doctor validates it now, the dashboard and the MCP tool included, and `check` stays as a deprecated alias for `site:doctor` in the current directory. Folding it in also fixed a worker it was wrong about: one declared in the project's own `custom_workers` was only recognised on a custom-container site, so a plain site running one was told it had no definition to match.

The reclaim offer the issue asks for is already there, doctor lists it as an optional fix and `--fix` runs `lerd cleanup` after confirming.

Closes #1358


geodro:
Ran the gate on this and it is green, build, vet, gofmt, the full suite and a race run over sitedoctor and cli. On a twenty site machine the sweep costs about two seconds, and the project config check shows up where it should, both in site:doctor and in the alias.

One thing to fix before this goes in. In a directory with no .lerd.yaml, lerd check now prints all checks pass and exits 0, where it used to fail with no .lerd.yaml found, run lerd init. Running it one directory too high used to say so, now it says the project is healthy. The alias should still bail when the file is not there.

Three smaller ones. The sweep line reads 2 warning, sitedoctor already has a plural helper for that. cachedDatabases holds one global mutex across the lookup so a postgres exec blocks a mysql one, which does not matter at this scale, but the comment says callers only wait per engine, so either the lock goes per service or the comment goes. And validPHPVersion is now the third copy of the same MAJOR.MINOR rule, next to the one in cli and the one in mcp, worth exporting one instead.

Checked what folding check in might have dropped, the service installed and db.service validation is covered by required_services, which resolves db.service too, so nothing is lost there.

Thanks for taking this one on, it is a good change.

PR #1493, opened 2026-08-13
Show what changed after a self-managed update lands

### LINKED ISSUE #1359: feat(update): show what changed once the update lands
lerd update checks, confirms, installs and refreshes the framework definitions the sites use, then stops without a word about what actually changed. doctor and status both end by telling the user to run lerd whatsnew, so the two commands that did not update anything advertise the changelog while the one that just did says nothing.

After a successful update it should show the same summary lerd whatsnew prints for the versions it moved across. A package managed install is deferred to apt, dnf or brew and never reaches that point, so this only covers the self managed path, and an unattended run wants it suppressed rather than dumped into a log.

JoeJoeflyn:
After a self-managed update lands, print the changelog that was already fetched for the confirmation prompt. Gated on feedback.Interactive() so unattended runs suppress it. Package-managed installs never reach this point.

Closes #1359


geodro:
Thanks for picking this one up, and sorry for the long reply, it turned out there is something broken underneath it.

Tried this on a real update in a VM, 1.33.0 to 1.33.1, and the new block never printed. FetchChangelog reads CHANGELOG.md at the repo root, which is a symlink to docs/changelog.md, and raw.githubusercontent serves the link target as the body, so the parser sees no version headings and comes back empty. That has been true since April, which is also why lerd whatsnew has been answering no changelog entries found. The block earlier in update falls back to the release URL, so nothing on screen ever gave it away. Not your doing, you just landed on top of it.

Two things worth changing while you are in there. It should read docs/changelog.md at the tag it just installed rather than at main, so what prints is the changelog of the version that landed and not whatever main happens to be holding. And the output wants to be a short list, the headline of each entry with its issue numbers, rather than the whole thing. I pointed the fetch at a working URL to watch it run, and the 1.33.1 changelog is around nine thousand characters, so it printed in full before the confirmation and then in full again at the end of the same run.

The empty case wants mirroring too, if the fetch returns nothing the earlier block prints the release link and this one prints nothing at all. And since the short list means parsing, it should come with a test.

JoeJoeflyn:
good catch on the symlink, sorry for missing it, updated

PR #1547, opened 2026-08-23
feat: register the lerd MCP server with OpenCode

### LINKED ISSUE #1542: Add OpenCode support to `lerd mcp:enable-global` and `lerd mcp:inject`
I'm setting up Lerd for the first time on a test machine (running Omarchy Quattro) to try it out, and so far it's waaaaaay more advanced and stable than I expected!! Amazing work, seriously!

But, I'm missing the MCP in my OpenCode sessions. I've added it manually, but it would be nice if Lerd could automate this, like everything else! 

Thanks again!

geodro:
OpenCode is the one client so far that shares neither of the two JSON shapes lerd already writes. Its entry lives under an mcp key rather than mcpServers or servers, names the transport local rather than stdio, folds the command and its arguments into a single array, and expects the entry to say it is enabled. That made it a new format rather than another row in the table, so lerdJSONEntry now takes the client it is building for instead of a bare needsType flag.

Everything else was already general enough to carry it. mergeServerJSON and removeServerJSON key off the client's own ServerKey, so the merge, the removal and the has-lerd check needed no changes at all.

OpenCode reads AGENTS.md from the project root, which the codex entry already writes, so the only context file it claims for itself is the copy under its own config directory. That is the same reasoning the antigravity entry uses for GEMINI.md.

Registered at opencode.json in the project root and ~/.config/opencode/opencode.json globally, both taken from OpenCode's own documentation rather than inferred.

Closes #1542


tjbarber:
Thank you for adding this!

PR #1811, opened 2026-09-15
fix: treat a machine that is already up as started

### LINKED ISSUE #1810: lerd start aborts when podman misreports the machine as not running
On macOS with podman 6.0.0, `lerd start` can abort before it starts nginx or any PHP FPM, leaving every site dead, because it trusts a machine state that podman itself contradicts one command later.

This happened twice in one day on lerd v1.35.0-beta.4. Both times the VM was genuinely healthy. `podman machine ssh --username root 'uptime'` reported nine and a half hours of uptime with a load average of 0.10, and containers started normally once the launchd units were bootstrapped by hand.

The state read is the problem. Right now, with containers running and sites serving over that same VM, podman reports this:

```
$ podman machine list --format "{{.Name}}\t{{.Running}}"
podman-machine-default	false
```

`ensurePodmanMachineRunning` reads exactly that format string, so `m.running` is false and the early return at the `else if m.running` branch never fires. Control falls through to `startPodmanMachineWithRetry`, and podman then contradicts its own listing:

```
Error: unable to start "podman-machine-default": already running
```

Exit 125. `startPodmanMachineWithRetry` treats that as a failed boot, waits three seconds, tries again, gets the same answer, and returns an error. `ensurePodmanMachineRunning`'s caller returns on that error, so nginx and the FPM pools are never touched and every `.test` site stays down.

The guidance printed at that point makes it worse. It suggests `podman machine stop && podman machine start`, and failing that `lerd machine reset` to recreate the VM followed by `lerd install`. For this failure that advice is inverted: the VM is fine, and resetting it would destroy a working machine and its images to fix a status flag.

The narrow fix is to treat "already running" from `podman machine start` as success rather than as a boot failure. That keeps the vfkit retry for the case it was written for, while making the start path resilient to a state read that cannot be trusted. It would also have turned both of today's incidents into no-ops, since the only thing actually missing was the launchd units, which `lerd start` goes on to handle.

For reference, eighteen of thirty two `lerd-*` units were absent from launchd when this happened, including nginx, mariadb, redis and every FPM pool. Why they detached is a separate question I have not answered yet.

prabowosd:
`podman machine list` can report a running machine as stopped, which is the only reason `ensurePodmanMachineRunning` attempts a start at all, and podman then refuses that start as "already running" with exit 125. `startPodmanMachineWithRetry` read that as a failed boot, so `lerd start` returned before touching nginx or any FPM pool and left every site down, while printing guidance that suggested resetting a VM which was healthy the whole time.

A failed start now asks `podman ps` whether container operations work before deciding anything. That exercises the whole stack rather than a status field, so it answers the question the failure actually leaves open, and it does not depend on how podman words its refusal. The vfkit retry is untouched and still runs when the VM genuinely has not booted.

The fake podman in the retry tests exited zero for every subcommand other than `machine start`, so `ps` appeared to work even when the VM had never booted. It now fails while the machine is down, which is what podman does.

Refs #1810


geodro:
@prabowosd thanks!

geodro (approved):

Issue #1842, opened 2026-09-17
lerd php:ports writes its PublishPort lines under [Install], so the extra FPM ports never get published


erhanurgun:
## Summary

Ports added with `lerd php:ports add` never reach the host. `ApplyExtraPorts` appends the `PublishPort=` lines to the end of the rendered FPM quadlet, but `lerd-php-fpm.container.tmpl` ends with `[Service]` and `[Install]`, so the lines land in `[Install]`. Quadlet only reads `PublishPort` from `[Container]`, so the container starts with no published ports while the command reports success. The macOS launchd writer builds its podman args from `[Container]` only as well, so it should drop them the same way, though I have only hit this on Linux.

This goes back to when `php:ports` shipped in 1.28.0 and is still the case on main. `TestFPMPortsRenderAsLoopbackPublish` only checks that the line is somewhere in the content, so it passes either way.

## Reproduce

```console
$ lerd php:ports add 8000 5273 --php 8.5
 → PHP 8.5: localhost:8000 -> container 8000
 → PHP 8.5: localhost:5273 -> container 5273
 ✓ shell ports updated for PHP 8.5
$ tail -6 ~/.config/containers/systemd/lerd-php85-fpm.container
[Install]
WantedBy=default.target
PublishPort=127.0.0.1:8000:8000
PublishPort=[::1]:8000:8000
PublishPort=127.0.0.1:5273:5273
PublishPort=[::1]:5273:5273
$ podman inspect lerd-php85-fpm --format '{{json .HostConfig.PortBindings}}'
{}
$ ss -ltn | grep -E ':(8000|5273)\b'
$
```

systemd logs why on the reload:

```
/run/user/1000/systemd/generator/lerd-php85-fpm.service:85: Unknown key 'PublishPort' in section [Install], ignoring.
```

The same four lines in a `[Container]` drop-in under `lerd-php85-fpm.container.d/` publish fine, and the dev server inside the container answers on localhost.

## Expected

`ApplyExtraPorts` should put the lines inside `[Container]`, the way `InjectPodmanArgs` places its line next to `Image=`, and the test should check the section rather than only the content.

## Environment

lerd 1.35.0 (8af835b), Podman 5.7.0, Ubuntu 26.04, rootless with systemd user units.

PR #1843, opened 2026-09-17
fix: publish php:ports mappings from the [Container] section

erhanurgun:
lerd php:ports add never published anything. ApplyExtraPorts appended the PublishPort lines to the end of the rendered FPM quadlet, and since the template ends with [Service] and [Install], they landed in [Install], where quadlet ignores them, so the container came up with no host ports while the command reported success.

The lines now go in right after Image=, the same place InjectPodmanArgs puts its line, so they stay in [Container] whatever sections the template grows after it. Both now share that insert.

Refs #1842


geodro (approved):
Thanks @erhanurgun

PR #1854, opened 2026-09-18
feat: ship the ODBC stack in the PHP images and register drivers with php:odbc

### LINKED ISSUE #1852: feat: ship the ODBC stack in the PHP images
Projects that talk to an ODBC data source (SAP HANA, Oracle, SQL Server, Snowflake) cannot use the lerd images as they are. There is no unixODBC in them and no odbc extensions, so an `odbc:` DSN has nothing to resolve through, and the way out is a per-site Containerfile that recompiles pdo_odbc on top of the base image, which every such project ends up working out for itself.

The bundle should carry ODBC the way it carries pdo_pgsql: unixODBC as the driver manager, ext-odbc for the `odbc_*` functions and their ini settings, and ext-pdo_odbc for the PDO path. Both need configure flags a rebuild on top of the image cannot pass, pdo_odbc picks the IBM DB2 backend otherwise and ext/odbc turns every backend on at once when built statically, so this belongs in the base image build where the gd and ftp flags already live.

The driver itself stays out, those are licensed downloads. What the image can do is be ready for one, which also means carrying gcompat and libstdc++, since every vendor driver is built against glibc while the images are Alpine/musl. The FrankenPHP image gets the same, so an Octane site is not a step behind.

### LINKED ISSUE #1853: feat: register ODBC drivers with lerd php:odbc
With the images carrying unixODBC, the step left to the user is the driver registry. Writing `/etc/odbcinst.ini` means a per-site Containerfile even when the driver needs nothing but a path, and it is per site, so the same driver is registered again for every project that uses it.

`lerd php:odbc add <name> <path>` should register a driver once, for every PHP version and every site: lerd generates the odbcinst registry and mounts it into the FPM, custom-FPM and FrankenPHP containers, and the driver file stays where the vendor put it. `list` and `remove` alongside it.

Registration should say what the image makes of the driver rather than only writing a file. unixODBC answers every failed dlopen at connect time as `Can't open lib '...' : file not found`, whatever the cause, so the three real causes have to be told apart when the driver is registered: the container cannot see the path, the image is missing a library, or every library resolved but musl does not carry a symbol the driver wants.

Nothing should be mounted until a driver is registered, so a site whose own image writes an odbcinst.ini keeps working untouched.

### LINKED ISSUE #1856: fix: make the php:odbc check prove something about the container that serves
`lerd php:odbc add` exists to tell the three ways a registration fails apart, because unixODBC reports all of them at connect time as `Can't open lib '...' : file not found`. The check it runs is a throwaway `podman run` that mounts the generated registry and the driver's directory and reads the driver back. That container is assembled by hand rather than from the mounts the PHP container actually gets, so what it proves is a statement about itself, not about the container that will serve the site.

Two consequences, both of which end with the command affirming a driver it has not established anything about.

The first is that the closing line does not distinguish "checked and fine" from "could not check". When the probe cannot run at all the error is swallowed as nothing to go on, and the affirmative line prints anyway:

```
$ PATH=/nonexistent lerd php:odbc add PgNoProbe /home/<user>/vendordrv/psqlodbcw.so
 → registering ODBC driver PgNoProbe
 ✓ driver PgNoProbe registered, use it as Driver={PgNoProbe} in a DSN
```

Nothing was inspected there. The same sentence appears when the probe has read the driver and found it sound, so the one case a user most needs to distinguish is the one the output hides. A line saying the driver was registered but not verified would cost nothing and would be true.

The second is structural. The probe decides for itself which directories to mount, while the quadlet decides separately, from the registered drivers, the parked directories and the home mount. Nothing keeps the two in agreement and nothing tests that they agree. They are consistent today, and they were briefly not during the work on #1854: a driver whose directory the quadlet had reason to leave out was still mounted by the probe, which read it happily and reported a driver the FPM container could not see as one that loads. That is the exact failure the feature was built to prevent, produced by the feature's own check.

Worth closing by having the check derive its mounts from the same place the quadlet does, so a driver the container will not be able to reach cannot pass, and by keeping the affirmative wording for the case where a driver was actually read back. Related to #1855, which is the other half of the probe being unreliable, and separate from it: there the probe runs and reads a mount SELinux has denied, here it either never runs or runs against mounts the real container will not have.

szabizs:
The images had no unixODBC and no odbc extensions, so a project talking to an ODBC data source could not use them as they were. Every such project ended up with a per-site Containerfile recompiling pdo_odbc on top of the base image, each one working it out again.

unixODBC is the driver manager in every image now, with ext-odbc for the odbc_* functions and their ini settings, and ext-pdo_odbc behind an odbc: DSN. Both needed configure flags that a rebuild on top of the image cannot pass, which is why they belong in the base build: pdo_odbc picks the IBM DB2 backend unless unixODBC is named, and ext/odbc built statically turns every backend on at once and stops at the first one it cannot find, so it gets a pass of its own with PHP_ALWAYS_SHARED neutered. gcompat and libstdc++ come along because every vendor driver is a glibc build while the images are Alpine/musl, and without them unixODBC cannot load one at all. The FrankenPHP image bakes the same two so an Octane site is not a step behind.

The drivers themselves stay out, those are licensed downloads, so the other half is registering one. lerd php:odbc add names a driver in a registry lerd generates and mounts into the FPM, custom-FPM and FrankenPHP containers, and the driver file stays where the vendor's installer put it. Registering reads the driver back from inside the image, because unixODBC answers every failed dlopen at connect time as "Can't open lib ... file not found" whatever the cause, and the three real causes want different answers: the container cannot see the path, the image is missing a library, or every library resolved but musl does not carry a symbol the driver wants. The SAP HANA client lands in the third case, and the symbols it names are exactly the shim it needs. Nothing is mounted until something is registered, so a site whose own image writes an odbcinst.ini keeps working untouched.

The PHP docs gained a section covering all of it, from getting the rebuilt image to a worked SAP HANA example for a client that needs patching before the loader will touch it.

Closes #1852
Closes #1853


geodro (approved):
Hey @szabizs thanks for the PR. I've added some fixes on some findings and works great now! Don't forget to star the repo if you enjoy the project!

szabizs:
Thanks @geodro ! I am using it with SAP Hana and it works great. Will commit further PR's with useful stuff that we use.

